PR #2, opened 2023-01-20
Add a basic Docker runner

agrare:
This runs a simple `docker run --rm [-e env=vars] image`

### DIFF
--- a/lib/manageiq/floe/workflow/runner.rb
+++ b/lib/manageiq/floe/workflow/runner.rb
@@ -1,0 +1,27 @@
+module ManageIQ
+  module Floe
+    class Workflow
+      class Runner
+        class << self
+          def for_resource(resource)
+            raise ArgumentError, "resource cannot be nil" if resource.nil?
+
+            scheme = resource.split("://").first
+
+            case scheme
+            when "docker"
+              # TODO detect if we should use Docker, Podman, or Kubernetes
+              Docker.new
+            else
+              raise "Invalid resource scheme [#{scheme}]"
+            end
+          end
+        end
+
+        def run!(image, env = {}, secrets = {})
+          raise NotImplementedError, "Must be implemented in a subclass"
+        end
+      end
+    end
+  end
+end
--- a/lib/manageiq/floe/workflow/runner/docker.rb
+++ b/lib/manageiq/floe/workflow/runner/docker.rb
@@ -1,0 +1,24 @@
+module ManageIQ
+  module Floe
+    class Workflow
+      class Runner
+        class Docker < ManageIQ::Floe::Workflow::Runner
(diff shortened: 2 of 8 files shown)


lib/manageiq/floe/workflow/runner/docker.rb
```diff
@@ -0,0 +1,24 @@
...
+  module Floe
+    class Workflow
+      class Runner
+        class Docker < ManageIQ::Floe::Workflow::Runner
+          def run!(resource, env = {}, _secrets = {})
+            raise ArgumentError, "Invalid resource" unless resource.start_with?("docker://")
+
+            image = resource.gsub("docker://", "")
+
+            params = ["run", ["--rm"]]
```
Fryguy, line 11:
```suggestion
            params = ["run", [:rm]]
```

lib/manageiq/floe/workflow/states/task.rb
```diff
@@ -19,6 +19,17 @@ def initialize(workflow, name, payload)
             @resource          = payload["Resource"]
             @timeout_seconds   = payload["TimeoutSeconds"]
           end
+
+          def run!
+            puts name
```
Fryguy, line 24: Fine for now - probably should go to a logger when we have it.

lib/manageiq/floe/workflow/runner.rb
```diff
@@ -0,0 +1,27 @@
+module ManageIQ
+  module Floe
+    class Workflow
+      class Runner
+        class << self
+          def for_resource(resource)
+            raise ArgumentError, "resource cannot be nil" if resource.nil?
+
+            scheme = resource.split("://").first
```
Fryguy, line 9: Probably should use URI.parse if we always expect a URI format.

Fryguy:
All very minor suggestions.

lib/manageiq/floe/workflow/runner.rb
agrare: :+1: done

lib/manageiq/floe/workflow/runner/docker.rb
agrare: :+1: Done

lib/manageiq/floe/workflow/runner.rb
agrare: I had to revert this because:
```
>> URI.parse('docker://hello-world:latest')
/usr/lib/ruby/3.0.0/uri/rfc3986_parser.rb:67:in `split': bad URI(is not URI?): "docker://hello-world:latest" (URI::InvalidURIError)
```

Also eventually we talked about supporting `arn:...` type resources so maybe we would `split(":")` in that case

PR #4, opened 2023-01-22
Add docker secrets

agrare:
NOTE this implementation uses docker volumes with a tmpfile and thus is not "secure" from other root users on the system, but docker is intended for development purposes only since podman or kubernetes will be used.

Future investigation to see if it is possible to pass a file descriptor (allowing for a hidden secret file) or some other mechanism tot pass docker secrets without requiring docker swarm and without requiring writing secrets to the filesystem.

### DIFF
--- a/lib/manageiq/floe/workflow/runner/docker.rb
+++ b/lib/manageiq/floe/workflow/runner/docker.rb
@@ -3,13 +3,22 @@ module ManageIQ
     class Workflow
       class Runner
         class Docker < ManageIQ::Floe::Workflow::Runner
-          def run!(resource, env = {}, _secrets = {})
-            raise ArgumentError, "Invalid resource" unless resource.start_with?("docker://")
+          def run!(resource, env = {}, secrets = {})
+            raise ArgumentError, "Invalid resource" unless resource&.start_with?("docker://")
 
             image = resource.gsub("docker://", "")
 
+            secrets_file = nil
+
+            if secrets && !secrets.empty?
+              require "tempfile"
+              secrets_file = Tempfile.new
+              secrets_file.write(secrets.to_json)
+            end
+
             params = ["run", :rm]
-            params += env.map { |k, v| [:e, "#{k}=#{v}"] } unless env.empty?
+            params += env.map { |k, v| [:e, "#{k}=#{v}"] } if env && !env.empty?
+            params << [:v, "#{secrets_file.path}:/run/secrets"] if secrets_file
             params << image
 
             require "awesome_spawn"
@@ -17,6 +26,8 @@ module ManageIQ
             result = AwesomeSpawn.run!("docker", :params => params)
 
             [result.exit_status, result.output]
+          ensure
+            File.unlink(secrets_file) if secrets_file
           end
         end
       end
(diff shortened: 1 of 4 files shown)


Fryguy:
Wondering if we should have a Swarm runner as well

lib/manageiq/floe/workflow/runner/docker.rb
```diff
@@ -3,20 +3,31 @@ module Floe
...
+            params << [:v, "#{secrets_file.path}:/run/secrets"] if secrets_file
             params << image
 
             require "awesome_spawn"
             logger.debug("Running docker: #{AwesomeSpawn.build_command_line("docker", params)}")
             result = AwesomeSpawn.run!("docker", :params => params)
 
             [result.exit_status, result.output]
+          ensure
+            File.unlink(secrets_file) if secrets_file
```
Fryguy, line 30: I never understood unlink vs delete (delete always seemed clearer to me)

spec/workflow/runner/docker_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    it "raises an exception for an invalid resource uri" do
+      expect { subject.run!("arn:abcd:efgh") }.to raise_error(ArgumentError, "Invalid resource")
+    end
+
+    it "calls docker run with the image name" do
+      expect(AwesomeSpawn)
+        .to receive(:run!)
+        .with("docker", :params => array_including("run", :rm, "hello-world:latest"))
+        .and_return(result)
```
Fryguy, line 20: There are built-in helpers in AwesomeSpawn for this (see https://github.com/ManageIQ/awesome_spawn/blob/master/lib/awesome_spawn/spec_helper.rb)

spec/workflow/runner/docker_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      subject.run!("docker://hello-world:latest", {"FOO" => "BAR"})
+    end
+
+    it "passes a secrets volume to docker run" do
+      expect(AwesomeSpawn)
+        .to receive(:run!)
+        .with("docker", :params => array_including("run", :rm, [:e, "FOO=BAR"], [:v, a_string_including(":/run/secrets")], "hello-world:latest"))
+        .and_return(result)
+
+      subject.run!("docker://hello-world:latest", {"FOO" => "BAR"}, {"luggage_password" => "1234"})
```
Fryguy, line 39: 😉 

```suggestion
      subject.run!("docker://hello-world:latest", {"FOO" => "BAR"}, {"luggage_password" => "12345"})
```

spec/workflow/runner/docker_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        .with("docker", :params => array_including("run", :rm, [:e, "FOO=BAR"], "hello-world:latest"))
+        .and_return(result)
+
+      subject.run!("docker://hello-world:latest", {"FOO" => "BAR"})
+    end
+
+    it "passes a secrets volume to docker run" do
+      expect(AwesomeSpawn)
+        .to receive(:run!)
+        .with("docker", :params => array_including("run", :rm, [:e, "FOO=BAR"], [:v, a_string_including(":/run/secrets")], "hello-world:latest"))
```
Fryguy, line 36: We should probably also add a spec to verify the contents of the tempfile (though finding the file might be hard)

spec/workflow/runner/docker_spec.rb
Fryguy: Implemented this in b9d598ad3c24876d225dd2e8f5a0a7c0e155d0ab

PR #6, opened 2023-01-27
Add Workflow.run! and Workflow.step

Fryguy:
This commit pushes executing a full workflow into the Workflow class. A notable change here is that the incoming context is not just the inputs, but the full context including "global", "states", and "current_state". This allows stepping, persisting the context, then resuming from that context.

@agrare I wanted to discuss this one first before merge, because it change the inputs and I didn't know if you had other thoughts on how this should be done.  In the future, I'm wondering if the incoming context should be a defined class instead of a raw Hash, since it's easy to start messing up the names.

Required for https://github.com/agrare/manageiq-providers-workflows/pull/1

### DIFF
--- a/exe/manageiq-floe
+++ b/exe/manageiq-floe
@@ -13,7 +13,4 @@ end
 require "logger"
 ManageIQ::Floe.logger = Logger.new(STDOUT)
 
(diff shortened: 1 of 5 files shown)


agrare (approved):
Love it looks great

PR #7, opened 2023-01-30
Implement pass state

agrare:
Use more_core_extensions' `Hash#store_path` with the `JsonPath.path` value to store results in a result path

### DIFF
--- a/lib/manageiq/floe/workflow/states/pass.rb
+++ b/lib/manageiq/floe/workflow/states/pass.rb
@@ -12,7 +12,7 @@ module ManageIQ
 
             @next        = payload["Next"]
             @result      = payload["Result"]
-            @result_path = payload["ResultPath"]
+            @result_path = JsonPath.new(payload["ResultPath"]) if payload.key?("ResultPath")
           end
         end
       end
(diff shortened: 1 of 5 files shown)


lib/manageiq/floe/workflow/states/pass.rb
```diff
@@ -8,11 +8,29 @@ class Pass < ManageIQ::Floe::Workflow::State
...
+            @result_path = JsonPath.new(payload["ResultPath"]) if payload.key?("ResultPath")
+          end
+
+          def run!
+            logger.info("Running state: [#{name}]")
+
+            if result
+              path = result_path.path[1..]
+                                .map { |v| v.match(/\[(?<name>.+)\]/)["name"] }
+                                .map { |v| v[0] == "'" ? v.gsub("'", "") : v.to_i }
```
agrare, line 26: This is gross and needs refactoring, but this is dealing with turning
```
>> JsonPath.new("$.store.book[0].author").path
=> ["$", "['store']", "['book']", "[0]", "['author']"]
```

Into
`["store", "book", 0, "author"]`

for `Hash#store_path`

Fryguy: ok yeah it's temporary, but should suffice for now.

spec/workflow/states/pass_spec.rb
```diff
@@ -0,0 +1,21 @@
+RSpec.describe ManageIQ::Floe::Workflow::States::Pass do
+  let(:workflow) { ManageIQ::Floe::Workflow.load(GEM_ROOT.join("examples/workflow.json")) }
```
Fryguy, line 2: I know I stole this for specs - ultimately we might want an independent one, but for now this is fine.

PR #9, opened 2023-01-31
Add Path, PayloadTemplate, and ReferencePath classes

agrare:


lib/manageiq/floe/workflow/path.rb
```diff
@@ -0,0 +1,31 @@
...
+          obj, path =
+            if payload.start_with?("$$")
+              [context, payload[1..]]
+            else
+              [input, payload]
+            end
+
+          results = JsonPath.on(obj, path)
+
+          results.count < 2 ? results.first : results
```
agrare, line 22: NOTE I'm not sure if there is a better way to handle this, but `JsonPath` returns an array even if a single element was matched.  This means we can't do a simple 
```
{"foo" => "bar"}
"$.foo" -> "bar"
```
it will end up as `"$.foo -> ["bar"]`

lib/manageiq/floe/workflow/payload_template.rb
```diff
@@ -0,0 +1,44 @@
...
+        attr_reader :payload, :context
+
+        def interpolate_value_nested(value, inputs)
+          case value
+          when Array
+            value.map { |val| interpolate_value_nested(val, inputs) }
+          when Hash
+            value.to_h do |key, val|
+              val = interpolate_value_nested(val, inputs)
+              key = key.gsub(/\.\$$/, "") if key.end_with?(".$")
```
agrare, line 27: NOTE we can't `key.gsub!` here because `key` is frozen coming from the hash we're looping over

PR #22, opened 2023-02-15
[WIP] Implement Kubernetes runner

agrare:


lib/manageiq/floe/workflow/runner/kubernetes.rb
```diff
@@ -3,6 +3,36 @@ module Floe
...
+          end
+
+          def run!(resource, env = {}, secrets = {})
+            raise ArgumentError, "Invalid resource" unless resource&.start_with?("docker://")
+
+            image = resource.gsub("docker://", "")
+
+            name = "#{image.split("/").last.split(":").first}-#{SecureRandom.uuid}"
+            params = ["run", :rm, :attach, [:image, image], [:restart, "Never"], name]
+            params += env.map { |k, v| [:env, "#{k}=#{v}"] } if env && !env.empty?
```
Fryguy, line 20: empty check isn't necessary - map will just return an empty array, and += will be a no-op

lib/manageiq/floe/workflow/runner/kubernetes.rb
```diff
@@ -3,6 +3,36 @@ module Floe
...
+            require "awesome_spawn"
+            require "securerandom"
+
+            super
+          end
+
+          def run!(resource, env = {}, secrets = {})
+            raise ArgumentError, "Invalid resource" unless resource&.start_with?("docker://")
+
+            image = resource.gsub("docker://", "")
```
Fryguy, line 16: Super-minor, but prefer `sub` over `gsub`, since it will stop after the first replacement

agrare: Both of these comments are pertinent to the existing runners so I opened a PR to fix the docker/podman runners first, https://github.com/agrare/manageiq-floe/pull/24

PR #23, opened 2023-02-15
Refactor base State#run! to simplify child classes

agrare:


lib/manageiq/floe/workflow/state.rb
```diff
@@ -42,9 +42,13 @@ def context
         def run!(input)
           logger.info("Running state: [#{name}] with input [#{input}]")
 
+          output = block_given? ? yield : input
```
Fryguy, line 45: If this returns `output, next_state`, then I think you can do `choice` here too.  Perhaps other methods would return nil fir next_state, and then the next line can be `next_state ||=` ?  Thoughts?

agrare: :+1: Done

PR #44, opened 2023-03-23
Add a Context class

agrare:
Add a `Context` class to handle the global Input/Output context object.  This should mirror what AWS Step Functions has with any additions that we require on top of that.

https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html

Step Functions Example:
```
{
    "Execution": {
        "Id": "arn:aws:states:us-east-1:123456789012:execution:stateMachineName:executionName",
        "Input": {
           "key": "value"
        },
        "Name": "executionName",
        "RoleArn": "arn:aws:iam::123456789012:role...",
        "StartTime": "2019-03-26T20:14:13.192Z"
    },
    "State": {
        "EnteredTime": "2019-03-26T20:14:13.192Z",
        "Name": "Test",
        "RetryCount": 3
    },
    "StateMachine": {
        "Id": "arn:aws:states:us-east-1:123456789012:stateMachine:stateMachineName",
        "Name": "stateMachineName"
    },
    "Task": {
        "Token": "h7XRiCdLtd/83p1E0dMccoxlzFhglsdkzpK9mBVKZsp7d9yrT1W"
    }
}
```


exe/floe
```diff
@@ -7,14 +7,15 @@ require "optimist"
 opts = Optimist.options do
   version "v#{Floe::VERSION}\n"
   opt :workflow, "Path to your workflow json", :type => :string, :required => true
-  opt :inputs, "JSON payload to input to the workflow", :type => :string, :default => '{}'
```
agrare, line 10: Not really related to this change but we changed this to `input` and `output` in the states, this should stay consistent

lib/floe/workflow/context.rb
```diff
@@ -0,0 +1,39 @@
...
+    class Context
+      def initialize(context = nil, input: {})
+        context = JSON.parse(context) if context&.kind_of?(String)
+
+        @context = context || {
+          "Execution"    => {
+            "Input" => input
+          },
+          "State"        => {},
+          "States"       => [],
```
agrare, line 14: NOTE `States` doesn't exist in the AWS Step Functions version but we need it to show the execution history.  I'm just appending the current `"State"` to the list of `"States"` after the step execution is finished.

lib/floe/workflow/context.rb
```diff
@@ -0,0 +1,39 @@
...
+        @context[key]
+      end
+
+      def []=(key, val)
+        @context[key] = val
+      end
+
+      def dig(*args)
+        @context.dig(*args)
+      end
```
agrare, line 32: I don't want this to just be free-for-all access to the `@context` object, maybe accessor functions like `#state` and `#execution` but we still need JsonPath access for e.g. `$$.State.EnteredTime`

lib/floe/workflow/states/task.rb
```diff
@@ -51,15 +51,16 @@ def retry!(retrier)
           return if retrier.nil?
 
           # If a different retrier is hit reset the context
-          if !context.key?("retrier") || context["retrier"]["error_equals"] != retrier.error_equals
-            context["retrier"] = {"error_equals" => retrier.error_equals, "retry_count" => 0}
+          if !context["State"].key?("RetryCount") || context["State"]["Retrier"] != retrier.error_equals
```
agrare, line 54: `["State"]["Retrier"]` also isn't in the step functions spec but they also say if a different `ErrorEquals` is matched you have to reset the RetryCount so we need context of _which_ `ErrorEquals` we're in

lib/floe/workflow.rb
```diff
@@ -5,40 +5,75 @@
...
+      @status = "running" if @status == "pending"
+      @context["Execution"]["StartTime"] ||= Time.now.utc
+
+      input = @context["State"]["Output"] || @context["Execution"]["Input"].dup
+
+      tick = Time.now.utc
+      @context["State"] = {
+        "EnteredTime" => tick,
+        "Name"        => current_state,
+        "Input"       => input
```
agrare, line 48: `"Input"` isn't part of the Step Functions spec

lib/floe/workflow.rb
```diff
@@ -5,40 +5,75 @@
...
+      }
+
+      next_state, output = current_state.run!(input)
+      tock = Time.now.utc
+
+      @context["State"]["FinishedTime"] = tock
+      @context["State"]["Duration"]     = tock - tick
+      @context["State"]["Output"]       = output
```
agrare, line 56: These three also aren't part of the step functions spec

exe/floe
```diff
@@ -33,6 +34,6 @@ runner_options = opts[:docker_runner_options].to_h { |opt| opt.split("=", 2) }
 
 Floe::Workflow::Runner.docker_runner = runner_klass.new(runner_options)
 
-output = workflow.run!
+workflow.run!
 
-puts output.inspect
+puts workflow.context.state["Output"].inspect
```
kbrock, line 39: grumble: I wish it had returned the output.
But I guess the `workflow` has always been stateful

agrare: Yeah I was surprised that `run` returned `self` but that is actually from Jason's commit https://github.com/ManageIQ/floe/pull/33/files#diff-7ab660aebba40fbc9105ec3f6fe80d20b0d0f5d368dc8a7c3db8294c1e91a077R64

lib/floe/workflow.rb
```diff
@@ -5,40 +5,67 @@
...
-      state = first_state
-      input = context.dup
+    def step
+      @status = "running" if @status == "pending"
+
+      input = @context["states"]&.last&.dig("output") || @context["global"]
+
+      tick = Time.now.utc
+      next_state, output = current_state.run!(input)
+      tock = Time.now.utc
```
kbrock, line 46: do we want to consider using

```ruby
Process.clock_gettime(Process::CLOCK_MONOTONIC)
```

via https://blog.dnsimple.com/2018/03/elapsed-time-with-ruby-the-right-way/

lib/floe/workflow/states/task.rb
Fryguy: Can we make PRs / suggestions to the spec?  Perhaps that should be there.

agrare: Well this isn't part of the states-language spec, they just say `This version of the States Language specification does not specify any contents of the Context Object.` [[ref]](https://states-language.net/#context-object) so I was going off of the [[step-functions documentation]](https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html) so there isn't a public "spec" as far as I know.

I figured if we provided at least what the aws-step-functions service provides then anything that was written with that in mind would work.

lib/floe/workflow.rb
```diff
@@ -5,40 +5,74 @@
...
+        "Duration"     => tock - tick,
+        "Output"       => output,
+        "Name"         => next_state&.name,
+        "Input"        => output
+      }
 
-      until state.nil?
-        state, output = state.run!(input)
-        input = output
+      @context["States"] << @context["State"]
```
kbrock, line 57: Think this needs to be a method on context

next_state = context.run_and_record(current_state)

agrare: Hm I'm not sure about `run_and_record` going through context since it is the state that we're running and context just keeps some info about the run.
Maybe we could move some of this into the run method, but it would look basically the same just be in a different place and we have one or two states which have to entirely override the run method right now so it'd be duplicated code.

agrare: Maybe the "return" that is set to `@context["State"]` could be returned by the `current_state.run!`?  I'll have to play with that a bit.

exe/floe
kbrock, line 39: Is `workflow#output` a global concept?

agrare: I think workflow output only makes sense once the workflow has ended, but yeah we could set a `workflow#output` once the state machine has ended

PR #47, opened 2023-06-02
Fix States::Wait Path initialize arguments

agrare:
The Path class was converted to take only a payload and the context is passed later, and the Wait state initializer wasn't updated to match.

This was changed in https://github.com/ManageIQ/floe/pull/27/files#diff-52717b57bdde93e7013b314f78b10e30611aa208e2f99d3ea0dabfda253fa2cbL13-R13 and I just missed the Wait state due to the missing spec coverage of this state (added)

### DIFF
--- a/spec/workflow/states/wait_spec.rb
+++ b/spec/workflow/states/wait_spec.rb
@@ -1,21 +1,22 @@
 RSpec.describe Floe::Workflow::States::Pass do
   let(:workflow) { Floe::Workflow.load(GEM_ROOT.join("examples/workflow.asl")) }
-  let(:state)    { workflow.states_by_name["PassState"] }
+  let(:state)    { workflow.states_by_name["WaitState"] }
 
   describe "#run!" do
-    it "sets the result to the result path" do
+    it "sleeps for the requested amount of time" do
+      expect(state).to receive(:sleep).with(state.seconds)
+
       _, output = state.run!({})
-      expect(output["result"]).to include(state.result)
     end
   end
 
   it "#to_dot" do
-    expect(state.to_dot).to eq "  PassState"
+    expect(state.to_dot).to eq "  WaitState"
   end
 
   it "#to_dot_transitions" do
     expect(state.to_dot_transitions).to eq [
-      "  PassState -> WaitState"
+      "  WaitState -> NextState"
     ]
   end
 end
--- a/examples/workflow.asl
+++ b/examples/workflow.asl
@@ -66,6 +66,12 @@
         "bar": "baz"
       },
       "ResultPath": "$.result",
+      "Next": "WaitState"
+    },
+
+    "WaitState": {
+      "Type": "Wait",
+      "Seconds": 1,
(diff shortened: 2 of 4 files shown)


spec/workflow/states/wait_spec.rb
```diff
@@ -0,0 +1,22 @@
+RSpec.describe Floe::Workflow::States::Pass do
+  let(:workflow) { Floe::Workflow.load(GEM_ROOT.join("examples/workflow.asl")) }
+  let(:state)    { workflow.states_by_name["WaitState"] }
+
+  describe "#run!" do
+    it "sets the result to the result path" do
+      allow_any_instance_of(Object).to receive(:sleep).with(state.seconds)
```
agrare, line 7: @Fryguy do you know a better way to stub `sleep()` ?  I could call it from a sleep instance method on the class but that feels like making an awkward change just for test stubbing

Fryguy: I think you can can stub sleep directly on the state, because technically it calls `self#sleep`, _then_ walks the inheritance chain.

```suggestion
      except(state).to receive(:sleep).with(state.seconds)
```

spec/workflow/states/wait_spec.rb
```diff
@@ -0,0 +1,22 @@
+RSpec.describe Floe::Workflow::States::Pass do
+  let(:workflow) { Floe::Workflow.load(GEM_ROOT.join("examples/workflow.asl")) }
+  let(:state)    { workflow.states_by_name["WaitState"] }
+
+  describe "#run!" do
+    it "sets the result to the result path" do
```
Fryguy, line 6: The test isn't showing this (or is this WIP?)

spec/workflow/states/wait_spec.rb
agrare: Oh yeah that does work, I assumed that would be looking for a `Floe::Wait#sleep` method

spec/workflow/states/wait_spec.rb
agrare: Fixed

spec/workflow/states/wait_spec.rb
```diff
@@ -0,0 +1,22 @@
+RSpec.describe Floe::Workflow::States::Pass do
+  let(:workflow) { Floe::Workflow.load(GEM_ROOT.join("examples/workflow.asl")) }
+  let(:state)    { workflow.states_by_name["WaitState"] }
+
+  describe "#run!" do
+    it "sleeps for the requested amount of time" do
+      allow(state).to receive(:sleep).with(state.seconds)
```
Fryguy, line 7: This we should do an `expect` here.  If we need it to _actually_ wait, then we can tack on `.and_call_original`

```suggestion
      expect(state).to receive(:sleep).with(state.seconds)
```

agrare: :+1: good catch, done

PR #48, opened 2023-06-20
Add options to Workflow::Runner initialize

agrare:
Add a way for the caller to initialize the docker runner and pass options.

This is going to be important for the kubernetes runner because we're going to have to pass options for authentication

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,11 +29,11 @@ bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}'
 ```
 
 By default Floe will use `docker` to run `docker://` type resources, but `podman` and `kubernetes` are also supported runners.
-A different runner can be specified with the `--runner` option:
+A different runner can be specified with the `--docker-runner` option:
 
 ```
-bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}' --docker_runner podman
-bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}' --docker_runner kubernetes --docker_runner-options namespace=default server=https://k8s.example.com:6443 token=my-token
+bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}' --docker-runner podman
+bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}' --docker-runner kubernetes --docker-runner-options namespace=default server=https://k8s.example.com:6443 token=my-token
 ```
 
 ### Ruby Library
--- a/README.md
+++ b/README.md
@@ -28,12 +28,12 @@ Floe can be run as a command-line utility or as a ruby class.
 bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}'
 ```
 
-By default Floe will use `docker` to run `docker://` type resources, but `podman` and `kubernet` are also supported runners.
+By default Floe will use `docker` to run `docker://` type resources, but `podman` and `kubernetes` are also supported runners.
 A different runner can be specified with the `--runner` option:
 
 ```
 bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}' --runner podman
-bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}' --runner kubernetes --runner-option namespace=default
+bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}' --runner kubernetes --runner-option namespace=default --runner-option server=https://k8s.example.com:6443 --runner-option token=my-token
 ```
 
 ### Ruby Library
@@ -51,7 +51,7 @@ require 'floe'
 
 Floe::Workflow::Runner.docker_runner = Floe::Workflow::Runner::Podman.new
 # Or
-Floe::Workflow::Runner.docker_runner = Floe::Workflow::Runner::Kubernetes.new("namespace" => "default")
+Floe::Workflow::Runner.docker_runner = Floe::Workflow::Runner::Kubernetes.new("namespace" => "default", "server" => "https://k8s.example.com:6443", "token" => "my-token")
 
 workflow = Floe::Workflow.load(File.read("workflow.asl"))
 workflow.run!
--- a/README.md
+++ b/README.md
@@ -28,10 +28,30 @@ Floe can be run as a command-line utility or as a ruby class.
 bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}'
 ```
 
+By default Floe will use `docker` to run `docker://` type resources, but `podman` and `kubernet` are also supported runners.
+A different runner can be specified with the `--runner` option:
+
+```
+bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}' --runner podman
+bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}' --runner kubernetes --runner-option namespace=default
+```
+
 ### Ruby Library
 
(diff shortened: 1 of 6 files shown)


exe/floe
```diff
@@ -7,15 +7,34 @@ require "optimist"
...
+  opt :runner, "Type of runner for docker images", :type => :string, :default => "docker"
+  opt :runner_option, "Options to pass to the runner", :type => :string, :multi => true
 end
 
 require "logger"
 Floe.logger = Logger.new(STDOUT)
 
 workflow = Floe::Workflow.load(opts[:workflow], opts[:inputs], opts[:credentials])
 
+if opts.key?(:runner)
```
Fryguy, line 21: Since you've specified a default then I don't thing there's ever a way for this not to be true.

README.md
```diff
@@ -28,6 +28,14 @@ Floe can be run as a command-line utility or as a ruby class.
 bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}'
 ```
 
+By default Floe will use `docker` to run `docker://` type resources, but `podman` and `kubernet` are also supported runners.
```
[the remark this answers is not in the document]
agrare: :+1:

exe/floe
```diff
@@ -7,15 +7,34 @@ require "optimist"
 opts = Optimist.options do
   version "v#{Floe::VERSION}\n"
   opt :workflow, "Path to your workflow json", :type => :string, :required => true
-  opt :inputs, "JSON payload to input to the workflow", :type => :string, :default => '{}'
-  opt :credentials, "JSON payload with credentials", :type => :string, :default => '{}'
+  opt :inputs, "JSON payload to input to the workflow", :type => :string, :default => "{}"
+  opt :credentials, "JSON payload with credentials", :type => :string, :default => "{}"
+  opt :runner, "Type of runner for docker images", :type => :string, :default => "docker"
+  opt :runner_option, "Options to pass to the runner", :type => :string, :multi => true
```
Fryguy, line 13: A better way to do this might be `:type => :strings`, and leave out the multi true.  In this mode, you specify each option with spaces, so the runner options would look like:

```diff
- --runner-option namespace=default --runner-option server=https://k8s.example.com:6443 --runner-option token=my-token
+ --runner-options namespace=default server=https://k8s.example.com:6443 token=my-token
```

agrare: Ah yeah that is better

exe/floe
```diff
@@ -7,15 +7,32 @@ require "optimist"
 opts = Optimist.options do
   version "v#{Floe::VERSION}\n"
   opt :workflow, "Path to your workflow json", :type => :string, :required => true
-  opt :inputs, "JSON payload to input to the workflow", :type => :string, :default => '{}'
-  opt :credentials, "JSON payload with credentials", :type => :string, :default => '{}'
+  opt :inputs, "JSON payload to input to the workflow", :type => :string, :default => "{}"
+  opt :credentials, "JSON payload with credentials", :type => :string, :default => "{}"
+  opt :runner, "Type of runner for docker images", :type => :string, :default => "docker"
+  opt :runner_option, "Options to pass to the runner", :type => :strings
```
Fryguy, line 13: Think this reads better (also you'll need to update the readme)
```suggestion
  opt :runner_options, "Options to pass to the runner", :type => :strings
```

exe/floe
```diff
@@ -7,15 +7,32 @@ require "optimist"
...
 
+runner_klass = case opts[:runner]
+               when "docker"
+                 Floe::Workflow::Runner::Docker
+               when "podman"
+                 Floe::Workflow::Runner::Podman
+               when "kubernetes"
+                 Floe::Workflow::Runner::Kubernetes
+               else
+                 raise
```
Fryguy, line 29: The way I usually handle this is up above, after the Optimist.options, I do some sanity checks on the incoming parameters.  So in this case, I would do something like:

```ruby
opts = Optimist.options do
  # ...
end
Optimist.die(:runner, "must be one of #{Floe::Workflow::Runner::TYPES.join(", ")}) unless Floe::Workflow::Runner::TYPES.include?(opts[:runner])
```

where `Floe::Workflow::Runner::TYPES` looks like `%[docker podman kubernetes]`

You could probably do a similar sanity check that each of the runner_options looks like the `x=y` pattern.

Fryguy: Optimist.die will give you a nicer error message and tell you to check --help and all that.

lib/floe/workflow/runner.rb
```diff
@@ -5,11 +5,14 @@ class Workflow
     class Runner
       include Logging
 
+      def initialize(_options = {})
+      end
+
       class << self
-        attr_writer :docker_runner_klass
+        attr_writer :docker_runner
```
Fryguy, line 12: Wonder if this should just be `runner` (i.e. remove the docker bit) - in the future we may not have docker-specific runners - for example, built-in methods or amazon lambdas.

Fryguy: ohhhh no, wait, this is specifically how to run the `docker://` scheme - this is fine.

README.md
```diff
@@ -28,6 +28,14 @@ Floe can be run as a command-line utility or as a ruby class.
 bundle exec ruby exe/floe --workflow examples/workflow.asl --inputs='{"foo": 1}'
 ```
 
+By default Floe will use `docker` to run `docker://` type resources, but `podman` and `kubernetes` are also supported runners.
+A different runner can be specified with the `--runner` option:
```
Fryguy, line 32: To match the internals - wondering if this should be --docker-runner and --docker-runner-options.  This way if we need runners for other types in the future there aren't conflicts.

lib/floe/workflow/runner.rb
agrare: It was actually because we might have non-docker resources like `lambda://` that I named this docker_runner instead of just runner.  We would have a runner class for each resource type that we support.

exe/floe
Fryguy, line 13: Minor for readability:

```suggestion
  opt :inputs, "JSON payload to input to the workflow", :type => :string, :default => "{}"
  opt :credentials, "JSON payload with credentials",    :type => :string, :default => "{}"
  opt :runner, "Type of runner for docker images",      :type => :string, :default => "docker"
  opt :runner_option, "Options to pass to the runner",  :type => :strings
```

super-technical, but if you specify a default, you don't need to specify the type because it's inferred from the default.  This could also be:

```suggestion
  opt :inputs, "JSON payload to input to the workflow", :default => "{}"
  opt :credentials, "JSON payload with credentials",    :default => "{}"
  opt :runner, "Type of runner for docker images",      :default => "docker"
  opt :runner_option, "Options to pass to the runner",  :type => :strings
```

exe/floe
```diff
@@ -6,16 +6,33 @@ require "optimist"
...
+runner_klass = case opts[:docker_runner]
+               when "docker"
+                 Floe::Workflow::Runner::Docker
+               when "podman"
+                 Floe::Workflow::Runner::Podman
+               when "kubernetes"
+                 Floe::Workflow::Runner::Kubernetes
+               end
+
+runner_options = opts[:docker_runner_options].to_h { |opt| opt.split("=") }
```
Fryguy, line 32: I believe we expect only the first `=` to be the delimiter, and any subsequent ones would be part of the value, so I think this is more accurate:

```suggestion
runner_options = opts[:docker_runner_options].to_h { |opt| opt.split("=", 2) }
```

Fryguy: Might be useful to have a spec for this case too

Fryguy: @agrare If possible, please add a spec for this case.

Fryguy: Oh nvm, I thought the runner spec was doing a full run using the CLI, but it's only stubbing the underlying kubernetes call and then calling fro the runner class directly.  Since we don't have specs for the `floe` script, I can merge now and we can include that in future specs for that executable.

PR #52, opened 2023-06-23
Add ability to run workflows asynchronously

agrare:
This adds the ability to run docker resources asynchronously by adding:
* `Floe::Workflow::Runner#run_async!`
* `Floe::Workflow::Runner#running?`
* `Floe::Workflow::Runner#success?`
* `Floe::Workflow::Runner#output`
* `Floe::Workflow::Runner#cleanup`

to the `Floe::Workflow::Runner` API

### DIFF
--- a/lib/floe/workflow/runner.rb
+++ b/lib/floe/workflow/runner.rb
@@ -30,7 +30,27 @@ module Floe
         end
       end
 
-      def run!(image, env = {}, secrets = {})
+      def run!(resource, env = {}, secrets = {})
+        raise NotImplementedError, "Must be implemented in a subclass"
+      end
+
+      def run_async!(_image, _env = {}, _secrets = {})
+        raise NotImplementedError, "Must be implemented in a subclass"
+      end
+
+      def running?(_ref)
+        raise NotImplementedError, "Must be implemented in a subclass"
+      end
+
+      def success?(_ref)
+        raise NotImplementedError, "Must be implemented in a subclass"
+      end
+
+      def output(_ref)
+        raise NotImplementedError, "Must be implemented in a subclass"
+      end
+
+      def cleanup(_ref, _secret)
         raise NotImplementedError, "Must be implemented in a subclass"
       end
     end
--- a/lib/floe/workflow/runner/docker.rb
+++ b/lib/floe/workflow/runner/docker.rb
@@ -43,9 +43,70 @@ module Floe
           secrets_file&.close!
         end
 
+        def run_async!(resource, env = {}, secrets = {})
+          raise ArgumentError, "Invalid resource" unless resource&.start_with?("docker://")
+
+          image = resource.sub("docker://", "")
+
+          params = ["run", :detach]
+          params += [[:net, "host"]] if network == "host"
+          params += env.map { |k, v| [:e, "#{k}=#{v}"] } if env
+
+          secrets_file = nil
+
+          if secrets && !secrets.empty?
+            secrets_file = Tempfile.new
+            secrets_file.write(secrets.to_json)
+            secrets_file.flush
+
+            params << [:e, "SECRETS=/run/secrets"]
+            params << [:v, "#{secrets_file.path}:/run/secrets:z"]
+          end
+
+          params << image
+
+          logger.debug("Running docker: #{AwesomeSpawn.build_command_line("docker", params)}")
+
+          begin
+            container_id = docker!(*params).output
+          rescue
+            secrets_file&.close!
+            raise
+          end
+
+          [container_id, secrets_file]
+        end
+
+        def cleanup(container_id, secrets_file)
+          delete_container(container_id)
+          secrets_file&.close!
+        end
+
+        def running?(container_id)
+          JSON.parse(docker!("inspect", container_id).output).first.dig("State", "Running")
+        end
+
+        def success?(container_id)
+          JSON.parse(docker!("inspect", container_id).output).first.dig("State", "ExitCode") == 0
+        end
+
+        def output(container_id)
+          docker!("logs", container_id).output
+        end
+
         private
 
         attr_reader :network
+
+        def delete_container(container_id)
+          docker!("rm", container_id)
+        rescue
+          nil
+        end
+
+        def docker!(*params, **kwargs)
+          AwesomeSpawn.run!("docker", :params => params, **kwargs)
+        end
       end
     end
   end
--- a/lib/floe/workflow/runner/docker.rb
+++ b/lib/floe/workflow/runner/docker.rb
@@ -18,27 +18,15 @@ module Floe
 
           image = resource.sub("docker://", "")
 
-          params  = ["run", :rm]
-          params += [[:net, "host"]] if network == "host"
-          params += env.map { |k, v| [:e, "#{k}=#{v}"] } if env
-
           secrets_file = nil
-
           if secrets && !secrets.empty?
-            secrets_file = Tempfile.new
-            secrets_file.write(secrets.to_json)
-            secrets_file.flush
-
-            params << [:e, "_CREDENTIALS=/run/secrets"]
-            params << [:v, "#{secrets_file.path}:/run/secrets:z"]
+            secrets_file = create_secret(secrets)
+            env["_CREDENTIALS"] = "/run/secrets"
           end
 
-          params << image
+          output = run_container(image, env, secrets_file)
 
-          logger.debug("Running docker: #{AwesomeSpawn.build_command_line("docker", params)}")
-          result = AwesomeSpawn.run!("docker", :params => params)
-
-          [result.exit_status, result.output]
+          [0, output]
         ensure
           secrets_file&.close!
         end
@@ -48,27 +36,14 @@ module Floe
 
           image = resource.sub("docker://", "")
 
-          params = ["run", :detach]
-          params += [[:net, "host"]] if network == "host"
-          params += env.map { |k, v| [:e, "#{k}=#{v}"] } if env
-
           secrets_file = nil
-
           if secrets && !secrets.empty?
-            secrets_file = Tempfile.new
-            secrets_file.write(secrets.to_json)
-            secrets_file.flush
-
-            params << [:e, "SECRETS=/run/secrets"]
-            params << [:v, "#{secrets_file.path}:/run/secrets:z"]
+            secrets_file = create_secret(secrets)
+            env["_CREDENTIALS"] = "/run/secrets"
           end
 
-          params << image
-
-          logger.debug("Running docker: #{AwesomeSpawn.build_command_line("docker", params)}")
-
           begin
-            container_id = docker!(*params).output
+            container_id = run_container(image, env, secrets_file, :detached => true)
           rescue
             cleanup(container_id, secrets_file)
             raise
@@ -98,6 +73,20 @@ module Floe
 
         attr_reader :network
 
+        def run_container(image, env, secrets_file, detached: false)
+          params  = ["run"]
+          params << (detached ? :detach : :rm)
+          params += env.map { |k, v| [:e, "#{k}=#{v}"] }
+          params << [:net, "host"] if @network == "host"
+          params << [:v, "#{secrets_file.path}:/run/secrets:z"] if secrets_file
+          params << image
+
+          logger.debug("Running docker: #{AwesomeSpawn.build_command_line("docker", params)}")
+
+          result = docker!(*params)
+          result.output
+        end
+
         def inspect_container(container_id)
           JSON.parse(docker!("inspect", container_id).output)
         end
@@ -106,6 +95,13 @@ module Floe
           docker!("rm", container_id)
         rescue
           nil
+        end
+
+        def create_secret(secrets)
+          secrets_file = Tempfile.new
+          secrets_file.write(secrets.to_json)
+          secrets_file.flush
+          secrets_file
         end
 
         def docker!(*params, **kwargs)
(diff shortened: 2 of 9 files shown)


Fryguy:
Does it help to have #6 ?  In my mind that helps with the full workflow run, particularly on the MIQ side where we will use the queue.

agrare:
> Does it help to have https://github.com/ManageIQ/floe/pull/6 ? In my mind that helps with the full workflow run, particularly on the MIQ side where we will use the queue.

Yes it does, @kbrock merged that recently

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -20,26 +20,49 @@ def initialize(options = {})
...
-          delete_secret!(secret) if secret
+        def cleanup(pod, secret)
+          delete_pod(pod)       if pod
+          delete_secret(secret) if secret
         end
 
         private
 
+        def pod_info(pod_name)
+          results = kubectl!("get", "pod", pod_name, [:output, :json], [:namespace, namespace])
```
Fryguy, line 62: Would a watch be better or is that overkill?

Fryguy: Oh we're calling kubectl directly - thought we were using something like kubeclient. 🤔

Fryguy:
I realize this is WIP, but it might be helpful to have specs where different commands could fail - i.e. stub_bad_run!

lib/floe/workflow/runner.rb
```diff
@@ -30,7 +30,36 @@ def for_resource(resource)
         end
       end
 
-      def run!(image, env = {}, secrets = {})
+      def run!(resource, env = {}, secrets = {})
+        ref, secret = run_async!(resource, env, secrets)
+
+        sleep(1) while running?(ref)
```
Fryguy, line 36: Should there by an upper limit on this?  (Maybe not for this PR, but perhaps we can make that configurable).

lib/floe/workflow/runner/kubernetes.rb
agrare: So I'm currently using `run` to call the async methods in order to confirm that async behaves the same as sync but it would be more efficient to leave `run` as a blocking call using fewer API calls.

For anything orchestrating >1 workflow run then a watch on all pods would absolutely be the way to go.

agrare:
Not directly related to this change, but rather than accept a `token=` argument on the command line use ENV var or `.kubeconfig`

lib/floe/workflow/states/task.rb
```diff
@@ -25,24 +25,24 @@ def initialize(workflow, name, payload)
           @credentials       = PayloadTemplate.new(payload["Credentials"])    if payload["Credentials"]
         end
 
-        def run!(*)
-          super do |input|
-            input = parameters.value(context, input) if parameters
+        def execute!(input)
+          input = input = input_path.value(context, input)
```
kbrock, line 29:
```suggestion
          input = input_path.value(context, input)
```

agrare: Ha thanks, replace all bites again

agrare:
Okay I pulled out some of the refactoring that this is based on since it wasn't directly related to asunc operation
https://github.com/ManageIQ/floe/pull/61

kbrock:
This is very cool.

an idea:

Can you write this PR so each runner implements synchronous `run!`.
So basically, you refactor the runners but not the workflows.

I feel that makes this completable.
From there refactoring the input / output / context may be more straight forward.

lib/floe/workflow/runner/podman.rb
```diff
@@ -11,17 +11,17 @@ def initialize(*)
...
+        def run_async!(resource, env = {}, secrets = {})
           raise ArgumentError, "Invalid resource" unless resource&.start_with?("docker://")
 
           image = resource.sub("docker://", "")
 
-          params = ["run", :rm]
+          params = ["run", :detach]
           params += env.map { |k, v| [:e, "#{k}=#{v}"] } if env
 
           if secrets && !secrets.empty?
```
kbrock, line 22: to make codeclimate happy, probably want to add something like `add_secrets`
Wasn't sure if `create_params` would make sense as well, but this seems simpler

```ruby
def add_secrets(params, secrets)
  return params unless secrets && !secrets.empty?

  ...

  params << [:e, "SECRETS=/run/secrets/#{secret_guid}"]
  params << [:secret, secret_guid]

  params
end
```

kbrock:
yea, this looks great

Pulling out a method to creates secrets will probably be trivial and get you up and running.

Maybe at a later time we can evaluate if credentials are a per task or whole workflow concept

lib/floe/workflow/runner/podman.rb
agrare: I wasn't too worried about codeclimate because the `run!` methods before were also over the cognitive complexity limit by the same amount so these aren't _really_ new (except the docker one)

lib/floe/workflow/runner.rb
```diff
@@ -30,7 +30,38 @@ def for_resource(resource)
         end
       end
 
-      def run!(image, env = {}, secrets = {})
+      def run!(resource, env = {}, secrets = {})
```
kbrock, line 33: call this `run_async_shim!`?

so we can keep run_async and run together?

kubernetes could call into the shim

lib/floe/workflow/runner/docker.rb
```diff
@@ -39,6 +39,68 @@ def run!(resource, env = {}, secrets = {})
         ensure
           secrets_file&.close!
         end
+
+        def run_async!(resource, env = {}, secrets = {})
+          raise ArgumentError, "Invalid resource" unless resource&.start_with?("docker://")
+
+          image = resource.sub("docker://", "")
+
+          params = ["run", :detach, [:net, "host"]]
```
Fryguy, line 48: As discussed, we should avoid `--net host`

 it will definitely require rewriting some of the examples, so I'm ok to merge for now, but then we need a followup issue to remove it.

agrare: We should discuss this separately since this isn't new code.
Docker isn't used for production only development and localhost:3000 is 99% of the development use cases so we can remove this if we want but it will require significant changes for developers' workflows

lib/floe/workflow/runner.rb
kbrock: the code lives on in kubernetes, so probably not necessary. we can address later

lib/floe/workflow/runner/docker.rb
Fryguy: Extracted to https://github.com/ManageIQ/floe/issues/77

lib/floe/workflow/runner.rb
```diff
@@ -30,7 +30,27 @@ def for_resource(resource)
...
+        raise NotImplementedError, "Must be implemented in a subclass"
+      end
+
+      def run_async!(_image, _env = {}, _secrets = {})
+        raise NotImplementedError, "Must be implemented in a subclass"
+      end
+
+      def running?(_ref)
```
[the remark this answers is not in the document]
agrare: This is the runner it doesn't have a workflow or a context reference.  This is checking if the external object (pod/container/lambda/etc...) is still running.  State#running? would be built on top of this.

lib/floe/workflow/runner/docker.rb
```diff
@@ -43,9 +43,70 @@ def run!(resource, env = {}, secrets = {})
...
+          end
+
+          params << image
+
+          logger.debug("Running docker: #{AwesomeSpawn.build_command_line("docker", params)}")
+
+          begin
+            container_id = docker!(*params).output
+          rescue
+            secrets_file&.close!
```
Fryguy, line 73: Should this call `cleanup` instead?  (and perhaps make `cleanup` "safer" in case the container failed to be created?

lib/floe/workflow/runner/docker.rb
```diff
@@ -43,9 +43,70 @@ def run!(resource, env = {}, secrets = {})
...
+          [container_id, secrets_file]
+        end
+
+        def cleanup(container_id, secrets_file)
+          delete_container(container_id)
+          secrets_file&.close!
+        end
+
+        def running?(container_id)
+          JSON.parse(docker!("inspect", container_id).output).first.dig("State", "Running")
```
Fryguy, line 86: Minori, but I think we should move `docker!("inspect", container_id).output` part to a separate method similar to how we have methods for other docker commands.  This way running? and success? can both use it (and the cod I think would be easier to read as there would be multiple parenthsis nestings)

spec/workflow/runner/docker_spec.rb
```diff
@@ -1,39 +1,101 @@
...
+      expect(secrets_file).to receive(:close!)
+      subject.cleanup(container_id, secrets_file)
     end
 
     context "with network=host" do
       let(:runner_options) { {"network" => "host"} }
 
       it "calls docker run with --net host" do
-        stub_good_run!("docker", :params => ["run", :rm, [:net, "host"], "hello-world:latest"])
+        stub_good_run!("docker", :params => ["run", :rm, "hello-world:latest"])
```
Fryguy, line 98: This change is confusing - the spec is specifically for when network -> host is being passed, so I would expect it on the command line.

Fryguy:
I noticed this kind of duplicates a lot from `run!`.  Is the plan to drop sync run, or DRY it up so they share more or just leave them completely separate?

agrare:
> I noticed this kind of duplicates a lot from run!. Is the plan to drop sync run, or DRY it up so they share more or just leave them completely separate?

I had done exactly this originally mainly to ensure that `run_async!` was passing all of the specs written for `run!`.  There are some inefficiencies implementing `run!` as `run_async! + wait` since it is essentially sleep in a loop until it is complete versus the external command exiting and returning immediately so I left them separate but I agree we need to move most of the logic to be as similar as possible which is what I've tried to do in the most recent commit.

spec/workflow/runner/docker_spec.rb
agrare: This looks like the result of a bad conflict resolution on my part, especially since it is testing `run!` in the `cleanup` describe block.

I've moved this back up to `describe run!` and fixed the missing `[:net, "host"]` (this was the result of redefining subject to not pass runner_options in the spec, not the actual code failing to pass `--net host`)

lib/floe/workflow/runner/docker.rb
```diff
@@ -18,34 +18,95 @@ def run!(resource, env = {}, secrets = {})
...
 
         attr_reader :network
+
+        def run_container(image, env, secrets_file, detached: false)
+          params  = ["run"]
+          params += detached ? [:detach] : [:rm]
+          params += [[:net, "host"]] if @network == "host"
+          params += env.map { |k, v| [:e, "#{k}=#{v}"] }
+          params << [:v, "#{secrets_file.path}:/run/secrets:z"] if secrets_file
+          params << image
```
Fryguy, line 82: For consistency, thinking we should use `<<` everywhere.

```suggestion
          params  = ["run"]
          params << detached ? :detach : :rm
          params << [:net, "host"] if @network == "host"
          params += env.map { |k, v| [:e, "#{k}=#{v}"] }
          params << [:v, "#{secrets_file.path}:/run/secrets:z"] if secrets_file
          params << image
```

lib/floe/workflow/runner/docker.rb
```diff
@@ -18,34 +18,95 @@ def run!(resource, env = {}, secrets = {})
...
+        end
+
+        def create_secret(secrets)
+          secrets_file = Tempfile.new
+          secrets_file.write(secrets.to_json)
+          secrets_file.flush
+          secrets_file
+        end
+
+        def docker!(*params, **kwargs)
```
Fryguy, line 107: Is `*params` correct?  Since we aren't splatting them below, and we are sending in a normal Array from the caller, I'm surprised this worked.

Fryguy: Oh, no actually, we are only sending a normal Array in one case, but it's that case I'm surprised works.

lib/floe/workflow/runner/docker.rb
agrare: I can `<<` the `--net host` one but
```
params << detached ? :detach : :rm
```

results in `false` not `:detach`
```
$ irb
>> detached = true
=> true
>> params = []
=> []
>> params << detached ? :detach : :rm
=> :detach
>> params
=> [true]
```

lib/floe/workflow/runner/docker.rb
agrare: The other callers are doing e.g. `docker!("rm", container_id)` it is only `run_container` that builds up an array of params because that one is more dynamic.  I liked the way `docker!("inspect", container)` looked more than `docker!(["inspect", container_id])` for most of the cases.  And we're passing in `docker!(*params)` which is why it works

lib/floe/workflow/runner/docker.rb
agrare: Done

lib/floe/workflow/runner/docker.rb
Fryguy: Oh my mistake...I misread the run! method and didn't see it splatted the array as in came in.

Fryguy (approved):

lib/floe/workflow/runner/docker.rb
Fryguy: Oh wild - I didn't realize that `<<` had priority over `?:`.

Only other thing would be the following, but it's up to you.

```ruby
params << (detached ? :detach : :rm)
```

agrare: Sure that works

agrare: Done

PR #53, opened 2023-06-26
Use kubeclient gem instead of kubectl

agrare:
Switch from using `kubectl` with `AwesomeSpawn` to the `kubeclient` gem.

Also adds support for `.kube/config` files.

### DIFF
--- a/lib/floe/workflow/runner/kubernetes.rb
+++ b/lib/floe/workflow/runner/kubernetes.rb
@@ -4,18 +4,51 @@ module Floe
   class Workflow
     class Runner
       class Kubernetes < Floe::Workflow::Runner
-        attr_reader :namespace, :server, :token
+        TOKEN_FILE   = "/run/secrets/kubernetes.io/serviceaccount/token"
+        CA_CERT_FILE = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 
         def initialize(options = {})
           require "awesome_spawn"
           require "securerandom"
           require "base64"
+          require "kubeclient"
           require "yaml"
 
-          @namespace = options.fetch("namespace", "default")
-          @server    = options.fetch("server", nil)
-          @token     = options.fetch("token", nil)
-          @token   ||= File.read(options["token_file"]) if options.key?("token_file")
+          kubeconfig_file    = ENV.fetch("KUBECONFIG", nil) || options.fetch("kubeconfig", File.join(Dir.home, ".kube", "config"))
+          kubeconfig_context = options["kubeconfig_context"]
+
+          kubeconfig  = Kubeclient::Config.read(kubeconfig_file) if File.exist?(kubeconfig_file)
+          kubecontext = kubeconfig.context(kubeconfig_context) if kubeconfig
+
+          token   = options["token"]
+          token ||= File.read(options["token_file"]) if options.key?("token_file")
+          token ||= File.read(TOKEN_FILE) if File.exist?(TOKEN_FILE)
+
+          server   = options["server"]
+          server ||= URI::HTTPS.build(:host => ENV.fetch("KUBERNETES_SERVICE_HOST"), :port => ENV.fetch("KUBERNETES_SERVICE_PORT", 6443)) if ENV.key?("KUBERNETES_SERVICE_HOST")
+
+          ca_file   = options["ca_file"]
+          ca_file ||= CA_CERT_FILE if File.exist?(CA_CERT_FILE)
+
+          verify_ssl = options["verify_ssl"] == "false" ? OpenSSL::SSL::VERIFY_NONE : OpenSSL::SSL::VERIFY_PEER
+
+          if server && token
+            ssl_options = {:verify_ssl => verify_ssl}
+            ssl_options[:ca_file] = ca_file if ca_file
+
+            auth_options = {:bearer_token => token}
+          elsif kubecontext
+            server       = kubecontext.api_endpoint
+            ssl_options  = kubecontext.ssl_options
+            auth_options = kubecontext.auth_options
+          else
+            raise ArgumentError, "Missing connections options, provide a kubeconfig file or pass server and token via --docker-runner-options"
+          end
+
+          @namespace  = options.fetch("namespace", "default")
+
+          @kubeclient = Kubeclient::Client.new(server, "v1", :ssl_options => ssl_options, :auth_options => auth_options)
+          @kubeclient.discover
 
           super
         end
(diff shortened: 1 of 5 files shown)


lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -47,8 +58,10 @@ def success?(pod_name)
         end
 
         def output(pod)
-          result = kubectl!("logs", pod, [:namespace, namespace])
-          result.output
+          # TODO no kubeclient method to get container logs
+          # URL: /api/v1/namespaces/:namespace/pods/:pod/log?container=:container
+
+          ""
```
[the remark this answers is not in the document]
agrare: Yes

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -4,42 +4,75 @@ module Floe
...
+          token     = options.fetch("token", nil)
+          token   ||= File.read(options["token_file"]) if options.key?("token_file")
+
+          kubeconfig_file = File.join(Dir.home, ".kube", "config")
+          kubeconfig      = Kubeclient::Config.read(kubeconfig_file) if File.exist?(kubeconfig_file)
+
+          api_endpoint = !kubeconfig.nil? ? kubeconfig.context.api_endpoint : options.fetch("server")
+          api_version  = options.fetch("api_version", "v1")
+          ssl_options  = !kubeconfig.nil? ? kubeconfig.context.ssl_options : {:ca_file => options.fetch("ca_file", nil), :verify_ssl => options.fetch("verify_ssl", OpenSSL::SSL::VERIFY_PEER)}
+          auth_options = !kubeconfig.nil? ? kubeconfig.context.auth_options : {:bearer_token => token}
```
Fryguy, line 28: Minor, but I think there should be rearranged slightly to group together the kubeconfig based things:

```suggestion
          if kubeconfig
            api_endpoint = kubeconfig.context.api_endpoint
            ssl_options  = kubeconfig.context.ssl_options
            auth_options = kubeconfig.context.auth_options
          else
            api_endpoint = options.fetch("server")
            ssl_options = {:ca_file => options.fetch("ca_file", nil), :verify_ssl => options.fetch("verify_ssl", OpenSSL::SSL::VERIFY_PEER)}
            auth_options = {:bearer_token => token}
          end
          api_version  = options.fetch("api_version", "v1")
```

agrare: I'm still playing around with this so expect this to change

agrare: Okay updated, what I wanted to achieve is always allowing the user to override the environment via command-line options so we prioritize e.g. `options.fetch("server")` over `kubeconfig.context.api_endpoint`, otherwise with the above proposal if you had a `kubeconfig` file you would never be able to connect to anything else.

So current logic is as follows:
1. Command-line options
2. `/run/secrets/kubernetes.io/serviceaccount/token` if it exists
3. `~/.kube/config` if it exists
4. Raise exception due to missing info

lib/floe/workflow/runner/kubernetes.rb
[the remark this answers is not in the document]
agrare: Fortunately I was looking in the wrong place and for the wrong string, kubeclient does have a method to get pod logs

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -4,42 +4,102 @@ module Floe
...
-          @server    = options.fetch("server", nil)
-          @token     = options.fetch("token", nil)
-          @token   ||= File.read(options["token_file"]) if options.key?("token_file")
+          kubeconfig_file    = ENV.fetch("KUBECONFIG", nil) || options.fetch("kubeconfig", File.join(Dir.home, ".kube", "config"))
+          kubeconfig_context = options.fetch("kubeconfig_context", nil)
+
+          kubeconfig  = Kubeclient::Config.read(kubeconfig_file) if File.exist?(kubeconfig_file)
+          kubecontext = kubeconfig.context(kubeconfig_context) if kubeconfig
+
+          token   = options.fetch("token", nil)
```
Fryguy, line 23: Since `nil` is the default, this might be less wordy?  (Same with other `options` calls below).  `fetch` makes some sense for ENV (and I think rubocop requests it), but it's not really useful for Hashes

```suggestion
          token   = options["token"]
```

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -51,72 +111,94 @@ def pod_name(image)
...
+            :metadata   => {
+              :name      => name,
+              :namespace => namespace
+            },
+            :spec       => {
+              :containers    => [
+                {
+                  :name  => container_name(image),
+                  :image => image,
+                  :env   => env.to_h.map { |k, v| {:name => k, :value => v.to_s} }
```
Fryguy, line 127: Is the `.to_h` necessary here?  The only thing I think it will ensure is that there aren't duplicate keys, but a .map should work the same on a Hash or an associative Array.

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -4,42 +4,102 @@ module Floe
...
+            server       = kubecontext.api_endpoint
+            ssl_options  = kubecontext.ssl_options
+            auth_options = kubecontext.auth_options
+          else
+            raise ArgumentError, "Missing connections options, provide a kubeconfig file or pass server and token via --docker-runner-options"
+          end
+
+          @namespace  = options.fetch("namespace", "default")
+
+          @kubeclient = Kubeclient::Client.new(server, "v1", :ssl_options => ssl_options, :auth_options => auth_options)
```
Fryguy, line 50: Not necessarily for this PR, but...

This creates a new client every time - I wonder if it would be helpful to cache the client and reuse it if the parameters are the same?  I'm guessing that in most scenarios you're not sending these steps to multiple clusters.

Fryguy:
Looks really good - some questions before merge.

lib/floe/workflow/runner/kubernetes.rb
agrare: So the docker_runner is only set once, I was actually worried about the opposite problem where there is a connection issue and we have to re-open the client.  I was considering caching the server, ssl_options, and auth_options and being able to `self.kubeclient = nil` then reopen it on a connection issue.

lib/floe/workflow/runner/kubernetes.rb
agrare: :+1: rubocop complained about that on `ENV` and I went overboard the other way :laughing:

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -4,42 +4,102 @@ module Floe
...
+        def running?(pod_name)
+          %w[Pending Running].include?(pod_info(pod_name).dig("status", "phase"))
+        end
+
+        def success?(pod_name)
+          pod_info(pod_name).dig("status", "phase") == "Succeeded"
+        end
+
+        def output(pod)
+          kubeclient.get_pod_log(pod, namespace).body
```
agrare, line 87: NOTE I think it is weird that `kubeclient` returns a `RestClient` object here instead of the body like the other calls.  Need to see if there was a reason for that or if it was an oversight.

Particularly because they move to Faraday in master so the next major release this would have to change.

PR #57, opened 2023-07-11
Set workflow output on completion

agrare:
When a workflow completes, set the last state's output as the workflow
output


agrare:
@kbrock this was one of the things we discussed last week in passing, lmk what you think

kbrock (commented):
@agrare LGTM. I had expected output to end up in the state (rather than an instance variable)
But this looks good

PR #58, opened 2023-07-11
Use the context accessor methods

agrare:
We have nice accessor methods for the context class so we should use them vs treating it like a hash


kbrock:
ok, thinking this needs to be rebased after #59

PR #59, opened 2023-07-11
CLOCK_MONOTONIC cannot be used to get wall time

agrare:
`CLOCK_MONOTONIC` cannot be used to get the wall time, it is only useful for comparing to other CLOCK_MONOTONIC values.

### DIFF
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -41,18 +41,19 @@ module Floe
 
       input = @context["State"]["Output"] || @context["Execution"]["Input"].dup
 
+      @context["State"] = {
+        "EnteredTime" => Time.now,
+        "Input"       => input,
+        "Name"        => current_state.name
+      }
+
       tick = Process.clock_gettime(Process::CLOCK_MONOTONIC)
       next_state, output = current_state.run!(input)
       tock = Process.clock_gettime(Process::CLOCK_MONOTONIC)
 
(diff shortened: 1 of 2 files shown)


lib/floe/workflow.rb
```diff
@@ -41,18 +41,19 @@ def step
...
-        "Duration"     => tock - tick,
-        "Output"       => output,
-        "Name"         => next_state&.name,
-        "Input"        => output
-      }
+      @context["State"]["FinishedTime"] = Time.now
+      @context["State"]["Duration"]     = (tock - tick) / 1_000_000.0
+      @context["State"]["Output"]       = output
 
       @context["States"] << @context["State"]
```
kbrock, line 58: should this be the `State` that we just built, or the previous state?

kbrock: do we need to save these when there is an error?

kbrock (approved):
LGTM

sorry about the suggestion that broke things

lib/floe/workflow.rb
agrare: > should this be the State that we just built, or the previous state?

This is the state that we're currently running

> do we need to save these when there is an error?

That is a good point I do need more robust error handling here.  The state run methods do handle exceptions already so they can execute the Fail conditions but if something is non-recoverable it might raise up here.

PR #61, opened 2023-07-13
Refactor input/output path handling

agrare:
Process input/output in child classes only since not all states have input_path or output_path properties.
I initially handled this in the base class to reduce duplication but since not all state types have this concept it is error prone and there isn't _that_ much duplication anyway

### DIFF
--- a/lib/floe/workflow/state.rb
+++ b/lib/floe/workflow/state.rb
@@ -45,27 +45,18 @@ module Floe
       def run!(input)
         logger.info("Running state: [#{name}] with input [#{input}]")
 
-        input = process_input(input)
-
         output     = execute!(input)
         next_state = workflow.states_by_name[@next] unless end?
-
-        output ||= input
-        output   = output_path&.value(context, output)
 
         logger.info("Running state: [#{name}] with input [#{input}]...Complete - next state: [#{next_state&.name}] output: [#{output}]")
 
         [next_state, output]
       end
 
-      def execute!(input)
-        input
-      end
-
       private
 
-      def process_input(input)
-        input_path.value(context, input)
+      def execute!(*)
+        raise NotImplementedError, "Must be implemented in a subclass"
       end
     end
   end
--- a/lib/floe/workflow/state.rb
+++ b/lib/floe/workflow/state.rb
@@ -45,9 +45,9 @@ module Floe
       def run!(input)
         logger.info("Running state: [#{name}] with input [#{input}]")
 
-        input = input_path.value(context, input)
+        input = process_input(input)
 
-        output     = block_given? ? yield(input) : input
+        output     = execute!(input)
         next_state = workflow.states_by_name[@next] unless end?
 
         output ||= input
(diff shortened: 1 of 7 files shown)


lib/floe/workflow.rb
```diff
@@ -60,10 +60,11 @@ def step
       context.states << context.state
 
       @status = current_state.status
-      @output = output if end?
 
       next_state_name = next_state&.name
-      @current_state = next_state_name && @states_by_name[next_state_name]
+      @current_state  = next_state_name && @states_by_name[next_state_name]
```
kbrock, line 65: think this is not a change

agrare: Was aligning the `=` up, but was from an unrelated change I'll pull into a separate PR

lib/floe/workflow/state.rb
```diff
@@ -45,18 +45,19 @@ def status
...
 
-        input = input_path.value(context, input)
-
-        output, next_state = block_given? ? yield(input) : input
-        next_state ||= workflow.states_by_name[payload["Next"]] unless end?
-
-        output ||= input
-        output   = output_path&.value(context, output)
+        output     = execute!(input)
+        next_state = workflow.states_by_name[@next] unless end?
```
kbrock, line 49: I like the stateless feel of `output = execute!(input)`. The feel that it is reading from the input and returning the output, and there aren't really any other side effects. But maybe `context` breaks this and it is an illusion. The `@next` is then jarring and shatters my nice stateless command pattern dreams.

I think of this like mouse routines where you have to convert the input to the local coordinate system and back to the global system.
`local` and `global` may be the wrong paradigm here, but this is what I'm thinking.

translate may be a better phrase, but I like the idea of the method name saying "I'm converting outside stuff to local stuff" and "I'm converting local stuff to outside stuff"


```ruby
  def localize_input(input)
    input_path ? input_path.value(context, input) : input
  end

  def globalize_next(next)
    workflow.states_by_name[next] if next && !end?
  end

  def globalize_output(output)
    output_path ? output_path.value(context, output) : output
  end

  # unsure if this is necessary, but feels like a parent thing
  def end?
    true
  end

  def run(input)
    next, output     = execute!(localize_input(input))
    [globalize_next(next), globalize_output(output)]
  end

```

It seems like `next` is pretty much hard coded except for `choice` (obvious why this is the case) and task (not as obvious but don't know what a task is and haven't been thinking too hard on it). So I can see why you wouldn't want to return from every state transition.

`end?` seems like it is overridden to be the same thing for all, so maybe moving that call into the states and removing that call from `globalize_next`/`State` may make sense.

kbrock: I feel like `run` and parent `State` should know about the other workflows, and `execute` is only worried about the nuances of the particular state node.

But this may be naïve since transitions like `choice` are all about choosing from the workflow provided values.

agrare:
Going to split this up into some separate prs

PR #62, opened 2023-07-14
Minor tweak to `Fail#end?`

kbrock:
`@end` is read from the ASL config file.
For `Fail`, this is also hard coded, but not in the ASL file.

After this, we hardcode the value


Fryguy:
@kbrock This method overriding also exists for some other states, so probably good to change them all.

agrare:
`Success` was the other state that overrode the `end?` method, https://github.com/ManageIQ/floe/pull/65

PR #63, opened 2023-07-14
[WIP] thoughts on parallel and async

kbrock:
This does not even qualify as a WIP.

I could go further, but felt I was a little beyond the quick ideas I wanted to jot down
This works best if the runners are async, but this should still be able to work if the runners are not async.

Think the active state needs to go into he context and replace current_state

### DIFF
--- a/exe/floe
+++ b/exe/floe
@@ -36,4 +36,4 @@ Floe::Workflow::Runner.docker_runner = runner_klass.new(runner_options)
 
 workflow.run!
 
-puts workflow.output.inspect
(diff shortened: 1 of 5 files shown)


kbrock:
update:
- fixed variable name typos

kbrock:
WIP: a lot of these concepts were extracted to other PRS

need to merge the next changes, map input output and the context changes.
Not sure if I will be able to salvage this

may need to merge the async changes first.

kbrock:
ideas are stale

PR #65, opened 2023-07-18
Remove `Success#end?` method override

agrare:
Use the `@end` instance variable rather than overriding the `#end?` method.

Follow-up to https://github.com/ManageIQ/floe/pull/62


kbrock:
yea. that parallel comment got me deep into the code and thinking.

PR #69, opened 2023-07-20
Fix double json.parse and context default value

agrare:
The Context initialize already handles parsing the input if it is a
string and fix the (unreachable) `||= {"global" => {}}` that would
return the wrong type of object.

### DIFF
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -18,12 +18,11 @@ module Floe
 
     def initialize(payload, context = nil, credentials = {})
       payload     = JSON.parse(payload)     if payload.kind_of?(String)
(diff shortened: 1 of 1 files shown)


lib/floe/workflow.rb
```diff
@@ -18,12 +18,11 @@ def load(path_or_io, context = nil, credentials = {})
 
     def initialize(payload, context = nil, credentials = {})
       payload     = JSON.parse(payload)     if payload.kind_of?(String)
-      context     = JSON.parse(context)     if context.kind_of?(String)
       credentials = JSON.parse(credentials) if credentials.kind_of?(String)
```
agrare, line 21: `Context#initialize` already does `JSON.parse` if the payload is a string

PR #70, opened 2023-07-20
Map input output

kbrock:
There is a concept of extracting data from the input context, and extracting data out of here for use in the output.

This formalizes the concept and moves into a parent class.

commits:

- [ ] https://github.com/ManageIQ/floe/pull/72 (first 5 commits)
- Introduce run_input that is basically run without the `next` part
- Introduce `MappedState` so states do not deal with mapping input and output

### DIFF
--- a/lib/floe/workflow/state.rb
+++ b/lib/floe/workflow/state.rb
@@ -27,6 +27,7 @@ module Floe
         @payload  = payload
         # All states must define Next or End except for Choice, Succeed, and Fail
         @end      = !!payload["End"]
+        @next     = payload["Next"]
         # All states must define a Type
         @type     = payload["Type"]
         # All states may define a Comment
@@ -44,7 +45,7 @@ module Floe
       end
 
       def status
-        end? ? "success" : "running"
+        "success"
       end
     end
   end
--- a/lib/floe/workflow/state.rb
+++ b/lib/floe/workflow/state.rb
@@ -40,6 +40,14 @@ module Floe
         @end
       end
 
+      def next_state
+        terminal_state? ? nil : @next
+      end
+
+      def run!(input)
+        [next_state, run_input!(input)]
+      end
+
       def context
         workflow.context
       end
--- a/lib/floe/workflow/state.rb
+++ b/lib/floe/workflow/state.rb
@@ -25,12 +25,17 @@ module Floe
         @workflow = workflow
         @name     = name
         @payload  = payload
+        # All states must define Next or End except for Choice, Succeed, and Fail
         @end      = !!payload["End"]
+        # All states must define a Type
         @type     = payload["Type"]
+        # All states may define a Comment
         @comment  = payload["Comment"]
       end
 
-      def end?
+      # https://states-language.net/#terminal-state
+      # @return [Boolean] true if there is no transition from this state
+      def terminal_state?
         @end
       end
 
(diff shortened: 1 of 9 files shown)


kbrock:
update:
- rebased onto converge_end_and_next
- can extract this if needed

lib/floe/workflow/states/task.rb
```diff
@@ -65,7 +64,6 @@ def retry!(retrier)
 
         def process_output!(output, results)
           return output if results.nil?
-          return if output_path.nil?
```
kbrock, line 68: Please verify that we want no output if output_path is nil.
Is it a configuration error, or a known use case.
I probably broke this behavior.

the other `input_path`/`output_path` implementation seemed to blow up in this case.

lib/floe/workflow/states/mapped_state.rb
```diff
@@ -0,0 +1,25 @@
...
+        def initialize(workflow, name, payload)
+          super
+          @input_path  = Path.new(payload.fetch("InputPath", "$"))
+          @output_path = Path.new(payload.fetch("OutputPath", "$"))
+        end
+
+        def run!(input)
+          input = input_path.value(context, input) if input_path
+          output = run_input(input)
+          output = output_path.value(context, output) if output_path
```
kbrock, line 18: Do we want to be lenient here?
Or is this a configuration issue and we should let it throw an error?

lib/floe/workflow/states/mapped_state.rb
```diff
@@ -0,0 +1,25 @@
...
+module Floe
+  class Workflow
+    module States
+      class MappedState < Floe::Workflow::State
+        attr_reader :input_path, :output_path
+
+        def initialize(workflow, name, payload)
+          super
+          @input_path  = Path.new(payload.fetch("InputPath", "$"))
+          @output_path = Path.new(payload.fetch("OutputPath", "$"))
```
kbrock, line 12: Larger question.

How do we want to handle invalid State payloads?

Do we want to throw `IllegalArgument` exceptions when the states are invalid?

lib/floe/workflow/state.rb
```diff
@@ -19,27 +19,40 @@ def build!(workflow, name, payload)
...
         @payload  = payload
-        @end      = !!payload["End"]
         @type     = payload["Type"]
         @comment  = payload["Comment"]
+        # some nodes like Fail, and Succeed have a hardcoded next of nil
+        # Choice ignores this (and has a default for their static fallback state)
+        @next     = payload["End"] ? nil : payload["Next"]
+        # all nodes but Fail use this:
+        @input_path  = Path.new(payload.fetch("InputPath", "$"))
+        @output_path = Path.new(payload.fetch("OutputPath", "$"))
```
agrare, line 35: The fact that Fail doesn't have `@input_path` and `@output_path` defined is why we didn't want to handle this in the base `State` class.  It doesn't make sense for `Fail` to have these ivars

lib/floe/workflow/state.rb
```diff
@@ -25,21 +25,42 @@ def initialize(workflow, name, payload)
         @workflow = workflow
         @name     = name
         @payload  = payload
+        # All states must define Next or End except for Choice, Succeed, and Fail
         @end      = !!payload["End"]
+        @next     = payload["Next"]
+        # All states must define a Type
         @type     = payload["Type"]
```
Fryguy, line 32: If we want to ensure it, then we can change this to the following (and then then comment is redundant)

```suggestion
        # All states must define a Type
        @type     = payload.fetch("Type")
```

kbrock:
will revist later

PR #71, opened 2023-07-20
Parse completed expressions rather than parse at runtime

kbrock:
Before
======

We create choice nodes at the top level
Then at lower levels, we create new choice nodes in runtime

After
=====

Create an AST for the complete choice expression

Thoughts
========

- Currently it is half creating the AST and half parsing at runtime. I'm ok with either but would prefer to stick with one or the other.
- We may feel that it is better to lazy parse these expressions rather than turn into a complete AST, since the JSON aspect is basically an AST.
-


kbrock:
update:
- fixed next variable (a couple of outstanding PRs overlap)

kbrock:
update:
- remove indentation change

kbrock:
update:
- extracted build_children (code climate complexity thing)

lib/floe/workflow/choice_rule.rb
```diff
@@ -4,24 +4,28 @@ module Floe
...
       end
 
-      attr_reader :next, :payload, :variable
+      attr_reader :next, :payload, :variable, :children
 
-      def initialize(payload)
-        @payload = payload
+      def initialize(payload, children = nil)
+        @payload   = payload
+        @children  = children
```
agrare, line 28: `children` defaults to `nil` here but it looks like if it actually were `nil` every subclass' `true?` call would fail doing `children.all?`, `children.first`, or `children.all?`

Should just we make `children` a required parameter?

lib/floe/workflow/choice_rule.rb
```diff
@@ -4,24 +4,28 @@ module Floe
   class Workflow
     class ChoiceRule
       class << self
-        def true?(payload, context, input)
-          build(payload).true?(context, input)
-        end
```
agrare, line 9: :+1: good catch looks like this was only used in specs

lib/floe/workflow/choice_rule.rb
```diff
@@ -4,24 +4,28 @@ module Floe
...
-
         def build(payload)
```
agrare, line 7: This is 100% NOT for this PR but it got me thinking we might want to add some validation here that the payload is valid per the rules defined in the spec [[ref]](https://states-language.net/#choice-state)

E.g.
> Boolean expression
>
> A Boolean Expression is a JSON object which contains a field named "And", "Or", or "Not". If the field name is "And" or "Or", the value MUST be an non-empty object array of Choice Rules, which MUST NOT contain "Next" fields; the interpreter processes the array elements in order, performing the boolean evaluations in the expected fashion, ceasing array processing when the boolean value has been unambiguously determined.
>
> The value of a Boolean Expression containing a "Not" field MUST be a single Choice Rule, that MUST NOT contain a "Next" field; it returns the inverse of the boolean to which the Choice Rule evaluates.

lib/floe/workflow/choice_rule.rb
kbrock: I think this was actually how everything was run.
all and/or/not ended up going this route and creating a rule for each stage.

lib/floe/workflow/choice_rule.rb
kbrock: yea. I was thinking the constructors could ensure that a next or end were defined among others.

and the boolean expression parsing could do the same

lib/floe/workflow/choice_rule.rb
kbrock: So I originally had `And`, `Or`, and `Not` extend from `Boolean`. The payload is basically the hash structure from the JSON. the children are the ast nodes (in this case, the `ChoiceRule` nodes)

lib/floe/workflow/choice_rule/boolean.rb
```diff
@@ -1,19 +0,0 @@
...
-    class ChoiceRule
-      class Boolean < Floe::Workflow::ChoiceRule
-        def true?(context, input)
-          if payload.key?("Not")
-            !ChoiceRule.true?(payload["Not"], context, input)
-          elsif payload.key?("And")
-            payload["And"].all? { |choice| ChoiceRule.true?(choice, context, input) }
-          else
-            payload["Or"].any? { |choice| ChoiceRule.true?(choice, context, input) }
-          end
```
kbrock, line 14: originally, this called the static `ChoiceRule.true?`. So every run created an AST node on the fly.

The new code creates these all up front.

advantages/disadvantages of each.
My main issue here is the Data nodes were creating a node up front, while these were creating them on the fly.
Prefer to have all choice rules follow a similar approach is possible

lib/floe/workflow/choice_rule.rb
agrare: Yeah I'd like to raise `Floe::InvalidWorkflowError` for things like this on initial load not at runtime so parsing this at the start is great

agrare (approved):

lib/floe/workflow/choice_rule.rb
```diff
@@ -4,24 +4,28 @@ module Floe
...
+            Floe::Workflow::ChoiceRule::And.new(payload, build_children(sub_payloads))
+          elsif (sub_payloads = payload["Or"])
+            Floe::Workflow::ChoiceRule::Or.new(payload, build_children(sub_payloads))
           else
-            Floe::Workflow::ChoiceRule::Boolean.new(payload)
+            Floe::Workflow::ChoiceRule::Data.new(payload)
           end
         end
+
+        def build_children(sub_payloads)
```
Fryguy, line 19: Super-minor, but this should probably be a private method.

PR #72, opened 2023-07-20
Converge end and next

kbrock:
All states have a "next state" concept:

- Next means there is a next state
- End means there is no next state / or the next state is to stop.

The commits are:

- separate `@end` (a tag) from `terminal_state?` (based upon tags and type)
- move `@next` up to `State` (think this was in one of your PRs)
- separate `@next` (tag) from `next_state` (based upon `@end`, `@next`, and `@type` ) 
- introduce `run_input!` which like `run!` but just focused on the input to output.

### DIFF
--- a/lib/floe/workflow/state.rb
+++ b/lib/floe/workflow/state.rb
@@ -27,6 +27,7 @@ module Floe
         @payload  = payload
         # All states must define Next or End except for Choice, Succeed, and Fail
         @end      = !!payload["End"]
+        @next     = payload["Next"]
         # All states must define a Type
         @type     = payload["Type"]
         # All states may define a Comment
@@ -44,7 +45,7 @@ module Floe
       end
 
       def status
-        end? ? "success" : "running"
+        "success"
       end
     end
   end
--- a/lib/floe/workflow/state.rb
+++ b/lib/floe/workflow/state.rb
@@ -40,6 +40,14 @@ module Floe
         @end
       end
 
+      def next_state
+        terminal_state? ? nil : @next
+      end
+
+      def run!(input)
+        [next_state, run_input!(input)]
+      end
+
       def context
         workflow.context
       end
--- a/lib/floe/workflow/state.rb
+++ b/lib/floe/workflow/state.rb
@@ -25,12 +25,17 @@ module Floe
         @workflow = workflow
         @name     = name
         @payload  = payload
+        # All states must define Next or End except for Choice, Succeed, and Fail
         @end      = !!payload["End"]
+        # All states must define a Type
         @type     = payload["Type"]
+        # All states may define a Comment
         @comment  = payload["Comment"]
       end
 
-      def end?
+      # https://states-language.net/#terminal-state
+      # @return [Boolean] true if there is no transition from this state
+      def terminal_state?
         @end
       end
 
--- a/spec/workflow/states/fail_spec.rb
+++ b/spec/workflow/states/fail_spec.rb
@@ -1,8 +1,14 @@
 RSpec.describe Floe::Workflow::States::Fail do
   let(:workflow) { Floe::Workflow.load(GEM_ROOT.join("examples/workflow.asl")) }
   let(:state)    { workflow.states_by_name["FailState"] }
+  let(:inputs)   { {} }
 
-  it "#end?" do
-    expect(state.end?).to be true
+  it "#terminal_state?" do
+    expect(state.terminal_state?).to be true
+  end
+
+  it "has a next of nil" do
+    next_state, _output = state.run!(inputs)
+    expect(next_state).to be nil
   end
 end
--- a/lib/floe/workflow/states/fail.rb
+++ b/lib/floe/workflow/states/fail.rb
@@ -13,8 +13,8 @@ module Floe
           @error = payload["Error"]
         end
 
-        def run!(input)
-          [nil, input]
+        def run_input!(input)
+          input
         end
 
         def terminal_state?
--- a/lib/floe/workflow/states/fail.rb
+++ b/lib/floe/workflow/states/fail.rb
@@ -11,11 +11,14 @@ module Floe
 
           @cause = payload["Cause"]
           @error = payload["Error"]
-          @end   = true
         end
 
         def run!(input)
           [nil, input]
+        end
+
+        def terminal_state?
+          true
         end
 
         def status
--- a/lib/floe/workflow/states/pass.rb
+++ b/lib/floe/workflow/states/pass.rb
@@ -17,12 +17,10 @@ module Floe
           @result_path = ReferencePath.new(payload.fetch("ResultPath", "$"))
         end
 
-        def run!(input)
+        def run_input!(input)
           output = input_path.value(context, input)
           output = result_path.set(output, result) if result && result_path
-          output = output_path.value(context, output)
-
-          [@next, output]
+          output_path.value(context, output)
         end
       end
     end
--- a/lib/floe/workflow/states/pass.rb
+++ b/lib/floe/workflow/states/pass.rb
@@ -4,12 +4,11 @@ module Floe
   class Workflow
     module States
       class Pass < Floe::Workflow::State
-        attr_reader :next, :result, :parameters, :input_path, :output_path, :result_path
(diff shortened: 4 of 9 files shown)


kbrock:
update:
- squashed 2 commits

kbrock:
update
- fixed spacing for `end_state` (rubocop)

lib/floe/workflow/catcher.rb
```diff
@@ -3,13 +3,13 @@
...
     class Catcher
-      attr_reader :error_equals, :next, :result_path
+      attr_reader :error_equals, :next_state, :result_path
 
       def initialize(payload)
         @payload = payload
 
         @error_equals = payload["ErrorEquals"]
-        @next         = payload["Next"]
+        @next_state   = payload["Next"]
```
Fryguy, line 12: The intention in the naming here, originally, was that the ivars matched the payload names to keep it consistent.

Though it does bring some clarity, I don't thin the name change is needed in this PR, unless I'm missing something.

spec/workflow/states/fail_spec.rb
```diff
@@ -1,8 +1,13 @@
...
   let(:workflow) { Floe::Workflow.load(GEM_ROOT.join("examples/workflow.asl")) }
   let(:state)    { workflow.states_by_name["FailState"] }
+  let(:inputs)   { {} }
 
   it "#end?" do
     expect(state.end?).to be true
   end
+
+  it "has a next of nil" do
+    expect(state.run!(inputs).first).to be nil
```
Fryguy, line 11: I would have expected a `.next` call here - not sure what .first is.

lib/floe/workflow/catcher.rb
agrare: Yeah I think having the name match the ASL property outweighs the keyword collision.  It is only an issue if you want to use the reader method instead of `@next` from within the class.

lib/floe/workflow/state.rb
```diff
@@ -19,27 +19,28 @@ def build!(workflow, name, payload)
...
+        @name       = name
+        @payload    = payload
+        end_state   = !!payload["End"]
+        @type       = payload["Type"]
+        @comment    = payload["Comment"]
+        @next_state = payload["Next"] unless end_state
       end
 
       def end?
-        @end
```
Fryguy, line 34: I think you're trying to remove the "end" ivar, but the reason we had it was because it's an actual value in the payload - so I'm not sure about removing it.  Personally I was trying to stick to that "if it's in the payload, we save the raw value in the ivar" to keep those concept separate, but perhaps that's too pedantic.

@agrare Thoughts?

This kind of started changing with https://github.com/ManageIQ/floe/pull/62 - that original code was intentional, because "End" doesn't exist in the payload, so it didn't seem right to hardcode an `@end` which to me implies it is in the payload.  If we want to move away from that pattern, I'm good with that, but want to make sure that it's the direction we want to go.

lib/floe/workflow/state.rb
```diff
@@ -19,27 +19,28 @@ def build!(workflow, name, payload)
...
-        @workflow = workflow
-        @name     = name
-        @payload  = payload
-        @end      = !!payload["End"]
-        @type     = payload["Type"]
-        @comment  = payload["Comment"]
+        @workflow   = workflow
+        @name       = name
+        @payload    = payload
+        end_state   = !!payload["End"]
```
Fryguy, line 28: Naming-wise this is a little weird to me, because end_state implies to me an actual state name or state object, but this is a boolean.  Perhaps

```suggestion
        is_end   = !!payload["End"]
```

lib/floe/workflow/state.rb
```diff
@@ -19,27 +19,28 @@ def build!(workflow, name, payload)
...
+        @type       = payload["Type"]
+        @comment    = payload["Comment"]
+        @next_state = payload["Next"] unless end_state
```
Fryguy, line 31: Or perhaps just change this to 

```suggestion
        @next_state = payload["Next"] unless payload["End"]
```

lib/floe/workflow/state.rb
agrare: Yeah I think this is a similar issue to the input_path/output_path issue, while it would be convenient to handle this in the base `State` class not every type of state has Next defined.  So having `@next` in the base `State` class breaks the pattern we've had of keeping very close to the ASL spec for our class definitions.

spec/workflow/states/fail_spec.rb
kbrock: `run` currently returns `[next, output]`, so it is the next part. I will make more clear

lib/floe/workflow/state.rb
kbrock: so I considered End and Next to be redundant.
But I appreciate making the two more similar

At a high level, all states have a `next`/`end?` concept, with `choice` and `task` being the only states that have a variable next.

If we want to go back to `@next` and `@end` and have each node parsing those concepts that works, too.

kbrock: as for input_path / output_path. having an intermediate node that handles path seemed like a good alternative.

kbrock:
ok, my driver for all of this is to extract the redundant `@next` all the states seem to have. Or the states like "fail" that have a hardcoded `nil` next, or those that accept an `end` option or a `next`.

I really liked the idea of `run_input` that just took input and produced output. And a few cases had mapping built on top of that run_input.

kbrock:
ok, going back over https://states-language.net/#state-type-table

All nodes except `Fail` have `InputPath, OutputPath`.
- These are optional for all.
- Having this concept defined in `State` makes sense to me.


All node except `Choice`, `Success`, and `Fail` have: `One of: Next or "End":true`.
- In my mind a single variable that stores the value could work.
- `Success` and `Fail` have an implicit `"End":true`.
- The concept of a static next state is required in all but `Choice`.
- ASIDE: `Choice` has one static next state called `default`, but this feels different.
- `Task`, `Parallel` and `Map` have an additional non-static `next` concept for retry/catch logic.
- Having the next concept defined in `State` makes sense to me.
- This state is an internally exposed concept and the only real interface is via run.

---

After Initially reading your comments, I started to tear out the `@end`. Keeping the stored state as close to the input is best. After looking at the table, I feel that these concepts are for all States.

I can see having helper methods handle these various concepts, and each state calling the methods accordingly. But I'm wondering if it is simpler to `run!` implement all of these behaviors including: the optional retry, catch, parameters, inputpath, outputpath. For nodes like success, they just won't be defined.

kbrock:
UNWIP: went back to the specs and I believe the `Next` and `End` concept are the same. (doesn't make sense to have heads:true, tails:true). Also I believe that the `Next` concept does belong in `State`.

update:
- applied `@next =` constructor changes per feedback. I am using `?:` instead of `unless` because I want to have `@next` always set explicitly.
- added run_input / run from 70. This is a building block for a number of future PRs and the main reason why `@next` was brought to the parent.

update:
- removed spacing / rubocop

agrare:
I think the biggest issue I have with this is that Next and End are actual fields defined for various states.

I think what you're trying to get to is asking "is this a terminal state"?  How about instead of trying to merge next and end, we leave those as simple fields from the payload, and add a `#terminal?` method.

https://states-language.net/#terminal-state
> The term "Terminal State" means a state with with { "End": true }, or a state with { "Type": "Succeed" }, or a state with { "Type": "Fail" }.

kbrock:
update:
- renamed end to terminal_state?
- added next_state used internally for `run!` (rather than overloading next

lib/floe/workflow/state.rb
```diff
@@ -25,21 +25,35 @@ def initialize(workflow, name, payload)
         @workflow = workflow
         @name     = name
         @payload  = payload
+        # All states must define Next or End except for Choice, Succeed, and Fail
         @end      = !!payload["End"]
+        @next     = payload["Next"]
+        # All states must define a Type
         @type     = payload["Type"]
+        # All states may define a Comment
         @comment  = payload["Comment"]
```
agrare, line 34: These are the only two properties that all states have
image

kbrock:
will revisit later

PR #78, opened 2023-08-02
Only transform payload-template values for .$ keys

agrare:
According to https://states-language.net/#payload-template values should only be transformed if the name ends with ".$".

> If any field within the Payload Template (however deeply nested) has a name ending with the characters ".$", its value is transformed according to rules below and the field is renamed to strip the ".$" suffix.


lib/floe/workflow/payload_template.rb
```diff
@@ -21,10 +21,11 @@ def interpolate_value_nested(value, context, inputs)
           value.map { |val| interpolate_value_nested(val, context, inputs) }
         when Hash
           value.to_h do |key, val|
-            val = interpolate_value_nested(val, context, inputs)
-            key = key.gsub(/\.\$$/, "") if key.end_with?(".$")
-
-            [key, val]
+            if key.end_with?(".$")
+              [key.gsub(/\.\$$/, ""), interpolate_value_nested(val, context, inputs)]
```
Fryguy, line 25: Better method for perf:

```suggestion
              [key.chomp(".$"), interpolate_value_nested(val, context, inputs)]
```

Fryguy:
Do we have a test that evaluates the case where the key ends in ".$"?  The only spec added here is when it _doesn't_ end it ".$"

agrare:
> Do we have a test that evaluates the case where the key ends in ".$"? The only spec added here is when it doesn't end it ".$"

There was already a test (look a few lines below) testing the case where it ends in `.$`

PR #80, opened 2023-08-02
move end out of the root state node

kbrock:
Goal is to have only valid parameters in each state node.

- Fixed the next state return values when end is nil
  Probably not necessary since next and end should not be defined in the same node.

---

Goal was to keep the code as close to the original, just moving `@end` into every node where it was valid and removing the `@end = true` since I think you said you didn't like that.

possibilities
- change `s/@end \? nil : @next/@next/g` -- both will not be defined on the same node.
- `s/end?/terminal_state?/g`

### DIFF
--- a/lib/floe/workflow/state.rb
+++ b/lib/floe/workflow/state.rb
@@ -25,21 +25,12 @@ module Floe
         @workflow = workflow
         @name     = name
         @payload  = payload
-        @end      = !!payload["End"]
         @type     = payload["Type"]
         @comment  = payload["Comment"]
       end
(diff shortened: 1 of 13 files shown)


kbrock:
update:
- put constructor back in for succeed. (in the future, input_path, output_path will go in there)

kbrock:
update:
- fix rubocop

kbrock:
update:
- add spec for `Wait#run!` and verifying `next_state` value

agrare (approved):

PR #82, opened 2023-08-07
Sleeping kubernetes

kbrock:
- in tests, don't sleep for a second, side step it
- reduce status checks on kubernetes pod

not tested (chuck if this deviates from the end goal API)


Fryguy:
I'm not sure why this is significantly different?

Fryguy:
I think your performance gains are strictly from changing the specs (i.e. you've changed every spec to do Running => Succeeded where before it was "Running => Running => Succeeded").  If you just change the specs and not the code do you still get the perf gains?

agrare:
Oh good catch this would have been introduced when we went to kubeclient.
I think you can just `expect(subject).to receive(:sleep).with(1)` in the kubernetes_spec though

Fryguy:
Oh I see - it's doing Running => Succeeded => Succeeded, which introduces a second unnecessary get_pod call.

Fryguy:
Regardless, this is clearer to me.  Should we also remove the `running?` and `success?` methods since they are no longer used?

kbrock:
udpate:
- added `with(1)` to sleep specs

Fryguy:
@kbrock Can you update the PR title to better reflect the change?

Fryguy (approved):

PR #83, opened 2023-08-07
Parse payload template at initialize not runtime

agrare:
Similar to https://github.com/ManageIQ/floe/pull/71

Parse PayloadTemplate values at workflow parse time rather than runtime.

### DIFF
--- a/lib/floe/workflow/payload_template.rb
+++ b/lib/floe/workflow/payload_template.rb
@@ -22,6 +22,10 @@ module Floe
         when Hash
           value.to_h do |key, val|
             if key.end_with?(".$")
+              if value.key?(key.chomp(".$"))
+                raise Floe::InvalidWorkflowError, "both #{key} and #{key.chomp(".$")} present"
+              end
+
               [key, parse_payload(val)]
             else
               [key, val]
--- a/lib/floe/workflow/payload_template.rb
+++ b/lib/floe/workflow/payload_template.rb
@@ -4,16 +4,17 @@ module Floe
   class Workflow
     class PayloadTemplate
       def initialize(payload)
-        @payload = parse_payload(payload)
+        @payload          = payload
+        @payload_template = parse_payload(payload)
       end
 
       def value(context, inputs = {})
-        interpolate_value(payload, context, inputs)
+        interpolate_value(payload_template, context, inputs)
       end
 
       private
 
-      attr_reader :payload
+      attr_reader :payload, :payload_template
 
       def parse_payload(value)
         case value
--- a/lib/floe/workflow/payload_template.rb
+++ b/lib/floe/workflow/payload_template.rb
@@ -18,47 +18,57 @@ module Floe
 
       def parse_payload(value)
         case value
-        when Array
-          value.map { |val| parse_payload(val) }
-        when Hash
-          value.to_h do |key, val|
-            if key.end_with?(".$")
-              if value.key?(key.chomp(".$"))
-                raise Floe::InvalidWorkflowError, "both #{key} and #{key.chomp(".$")} present"
-              end
-
-              [key, parse_payload(val)]
-            else
-              [key, val]
-            end
-          end
-        when String
-          if value.start_with?("$")
-            Path.new(value)
-          else
-            value
-          end
+        when Array  then parse_payload_array(value)
+        when Hash   then parse_payload_hash(value)
+        when String then parse_payload_string(value)
         else
           value
         end
       end
 
+      def parse_payload_array(value)
+        value.map { |val| parse_payload(val) }
+      end
+
+      def parse_payload_hash(value)
+        value.to_h do |key, val|
+          if key.end_with?(".$")
+            if value.key?(key.chomp(".$"))
+              raise Floe::InvalidWorkflowError, "both #{key} and #{key.chomp(".$")} present"
+            end
+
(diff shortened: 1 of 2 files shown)


lib/floe/workflow/payload_template.rb
```diff
@@ -4,31 +4,56 @@ module Floe
   class Workflow
     class PayloadTemplate
       def initialize(payload)
-        @payload = payload
+        @payload = parse_payload(payload)
```
kbrock, line 7: I think of `payload` as the unparsed input.
(I also kinda of want to get rid of all unparsed payloads that we store)

could we call this something like `parameters` or `template`?

lib/floe/workflow/payload_template.rb
```diff
@@ -4,31 +4,56 @@ module Floe
...
-        interpolate_value_nested(payload, context, inputs)
+        interpolate_value(payload, context, inputs)
       end
 
       private
 
       attr_reader :payload
 
-      def interpolate_value_nested(value, context, inputs)
+      def parse_payload(value)
```
kbrock, line 18: so cool to get `context` out of the parsing function. 🎉

lib/floe/workflow/payload_template.rb
```diff
@@ -4,31 +4,56 @@ module Floe
...
+            Path.new(value)
+          else
+            value
+          end
+        else
+          value
+        end
+      end
+
+      def interpolate_value(value, context, inputs)
```
kbrock, line 43: this works as you described it. And maybe we want to merge this as is and plan a possible continuation?

Does it make sense to take it a little further and have `PayloadTemplateArray`, `PayloadTemplateHash`, `PayloadTemplateStatic`, and `Path`?

Then all methods are like a command pattern with:

```ruby
def value(context, inputs)
  @template.value(context, inputs)
end
```

Maybe the recursive nature breaks this down too quickly.
Think I do like keeping `Path`, `.$` and static values hidden from the array/hash/constant parser.

Not sure if this is overcomplicating things though.

spec/workflow/payload_template_spec.rb
```diff
@@ -19,6 +19,24 @@
       it "returns the value from the inputs" do
         expect(subject.value(context, inputs)).to eq({"foo" => "bar", "bar" => "baz"})
       end
+
+      context "with an invalid payload" do
+        let(:payload) { {"foo.$" => "$.foo", "foo" => "bar"} }
+
+        it "raises an exception" do
+          expect { subject }.to raise_error(Floe::InvalidWorkflowError, "both foo.$ and foo present")
```
kbrock, line 27: nice

kbrock (commented):
This is going to be very nice.
Just those 2 comments

lib/floe/workflow/payload_template.rb
agrare: 👍 yeah good point

spec/workflow/payload_template_spec.rb
agrare: Yeah I noticed this in the spec and realized it was a super easy validation to add now

Fryguy (approved):

lib/floe/workflow/payload_template.rb
```diff
@@ -4,35 +4,73 @@ module Floe
   class Workflow
     class PayloadTemplate
       def initialize(payload)
-        @payload = payload
+        @payload          = payload
```
Fryguy, line 7: Hard to tell from this PR, but do we even need `@payload` anymore,  or is just having `@payload_template` sufficient?

agrare: It isn't used, we could remove it I just had it for consistency with the other classes which keep the raw payload

agrare: Removed

Fryguy:
I agree with the bot to avoid pinging `@payload` in the commit message (poor guy)

lib/floe/workflow/payload_template.rb
kbrock: Yay. I want to remove payload (and workflow) from everywhere. But more of a side thought and don't have any good push behind it

lib/floe/workflow/payload_template.rb
```diff
@@ -4,35 +4,85 @@ module Floe
   class Workflow
     class PayloadTemplate
       def initialize(payload)
-        @payload = payload
+        verify_payload!(payload)
```
kbrock, line 7: can we incorporate this in `parse_payload`?
Or do you feel it works better if you have separate?

I feel like an `else` or something in `parse` may meet our needs

agrare: :laughing: I moved this out of there to appease codeclimate.  I actually agree it was a little simpler in `parse_payload` but that cognitive complexity was too high

spec/workflow/payload_template_spec.rb
kbrock: aah. here is a good example of where a separate validation step is probably better than a parse step. But still curious your take.

lib/floe/workflow/payload_template.rb
agrare: Okay you convinced me to not change code just to fix a 6/5 cyclomatic complexity

Issue #64, opened 2023-07-14
Support intrinsic functions


kbrock:
There are 2 dozen intrinsic functions built into the language.
Would be nice to support them

https://states-language.net/#appendix-b

- [x] States.Format (https://github.com/ManageIQ/floe/pull/258)
- [x] States.StringToJson (https://github.com/ManageIQ/floe/pull/256)
- [x] States.JsonToString (https://github.com/ManageIQ/floe/pull/256)
- [x] States.Array (https://github.com/ManageIQ/floe/pull/194)
- [x] States.ArrayPartition (~~https://github.com/ManageIQ/floe/pull/239~~ https://github.com/ManageIQ/floe/pull/242)
- [x] States.ArrayContains (https://github.com/ManageIQ/floe/pull/242)
- [x] States.ArrayRange (https://github.com/ManageIQ/floe/pull/242)
- [x] States.ArrayGetItem (https://github.com/ManageIQ/floe/pull/242)
- [x] States.ArrayLength (https://github.com/ManageIQ/floe/pull/242)
- [x] States.ArrayUnique (https://github.com/ManageIQ/floe/pull/242)
- [x] States.Base64Encode (https://github.com/ManageIQ/floe/pull/242)
- [x] States.Base64Decode (https://github.com/ManageIQ/floe/pull/242)
- [x] States.Hash (https://github.com/ManageIQ/floe/pull/242)
- [x] States.JsonMerge (https://github.com/ManageIQ/floe/pull/255)
- [x] States.MathRandom (https://github.com/ManageIQ/floe/pull/242)
- [x] States.MathAdd (https://github.com/ManageIQ/floe/pull/242)
- [x] States.StringSplit (https://github.com/ManageIQ/floe/pull/242)
- [x] States.UUID (https://github.com/ManageIQ/floe/pull/194)

Fryguy:
States.Array and States.UUID completed in #194 - I'm updating the OP to enumerate the rest.

Fryguy:
Finished most of them in https://github.com/ManageIQ/floe/pull/242 - the ones left are more complicated, so going to do those separately.

Fryguy:
All done 🎉 

Thanks @kbrock for helping get this over the finish line :)

Fryguy:
Done - All released in 0.13.0

PR #84, opened 2023-08-08
[WIP] Add Intrinsic Functions

agrare:
Add the basics for Intrinsic Function support and implement StringToJson

Closes https://github.com/ManageIQ/floe/issues/64


lib/floe/workflow/intrinsic_function.rb
```diff
@@ -0,0 +1,56 @@
...
+  class Workflow
+    class IntrinsicFunction
+      class << self
+        def klass(payload)
+          function_module, function =
+            payload.match(/^(?<module>\w+)\.(?<function>\w+)\(.*\)$/)
+              .named_captures.values_at("module", "function")
+
+          case "#{function_module}.#{function}"
+          when "States.StringToJson"
```
agrare, line 13: NOTE trying to avoid a `const_get` here

Fryguy: Is there a specific reason why?  We can limit the const_get to just the class we care about if that's the concern.  (e.g. `Floe::Workflow::IntrinsicFunctions.const_get(function_module).const_get(function)`

lib/floe/workflow/intrinsic_function.rb
```diff
@@ -0,0 +1,55 @@
...
+          new(payload).value(context, input)
+        end
+
+        private alias_method :orig_new, :new
+
+        def new(*args)
+          if self == Floe::Workflow::IntrinsicFunction
+            klass(*args).new(*args)
+          else
+            orig_new(*args)
```
Fryguy, line 30: I think there's a slightly nicer way to do this "intanstiate from the abstract class" pattern without needing to alias_method, but I need to find how we did it elsewhere (Not going to hold up the PR for it)

agrare: :+1: I pulled this from LinuxAdmin FWIW

lib/floe/workflow/intrinsic_function.rb
```diff
@@ -0,0 +1,55 @@
...
+          end
+        end
+      end
+
+      attr_reader :args
+
+      def initialize(payload)
+        args  = payload.match(/^\w+\.\w+\((.*)\)$/).captures.first
```
Fryguy, line 38: Minor, but we might want to move this regex out to a constant and reuse on line 42

lib/floe/workflow/intrinsic_functions/states/string_to_json.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module Floe
+  class Workflow
+    module IntrinsicFunctions
+      module States
+        class StringToJson < Floe::Workflow::IntrinsicFunction
+          def value(context, inputs)
+            arg = args.first
+            arg =
+              if arg.kind_of?(Floe::Workflow::Path) || arg.kind_of?(Floe::Workflow::IntrinsicFunction)
```
Fryguy, line 11: Would it be cleaner to say `arg.respond_to?(:value)`?  Then if we have future evauable things they would just work.  That makes me wonder if it might be nice to have an Evaulable mixin, but I'm not sure what exactly we'd put into it.  I am seeing a lot of the same pattern of `if arg.start_with?("$.")` all over the place, which gets me wondering if there's way to DRY that up.

lib/floe/workflow/intrinsic_function.rb
Fryguy: Actually, is this the same regex as line 9 as well?

lib/floe/workflow/intrinsic_function.rb
Fryguy: haha - that's where I was gonna check first :D - let me check in a couple other places, but this is probably fine.

lib/floe/workflow/intrinsic_function.rb
Fryguy: No, I think it's ok, particularly when you are trying to "find" a subclass

lib/floe/workflow/intrinsic_function.rb
agrare: Done

lib/floe/workflow/intrinsic_function.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module Floe
+  class Workflow
+    class IntrinsicFunction
+      INTRINSIC_FUNCTION_REGEX = /^(?<module>\w+)\.(?<function>\w+)\((?<args>.*)\)$/.freeze
+
+      class << self
+        def klass(payload)
```
Fryguy, line 9: Minor, but I think it's clearer to describe the name with something like

```suggestion
        def detect_class(payload)
```

lib/floe/workflow/intrinsic_function.rb
```diff
@@ -0,0 +1,56 @@
...
+          new(payload).value(context, input)
+        end
+
+        private alias_method :orig_new, :new
+
+        def new(*args)
+          if self == Floe::Workflow::IntrinsicFunction
+            klass(*args).new(*args)
+          else
+            orig_new(*args)
```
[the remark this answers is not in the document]
agrare: Nice that worked perfectly thanks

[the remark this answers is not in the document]
Fryguy: I should make a PR over in LinuxAdmin too

lib/floe/workflow/intrinsic_function.rb
```diff
@@ -0,0 +1,56 @@
...
+          new(payload).value(context, input)
+        end
+
+        def new(*args)
+          if self == Floe::Workflow::IntrinsicFunction
+            detect_class(*args).new(*args)
+          else
+            super
+          end
+        end
```
Fryguy, line 33: Extremely, minor, but can we swap the order of these 2 methods?  I feel like ".new" is "more important" than ".value" and also I feel it should live next to detect_class.  Also, I'm wondering if detect_class should be marked private and perhaps listed _after_ ".new" (giving a final method ordering of "new", "detect_class", "value")

lib/floe/workflow/intrinsic_function.rb
```diff
@@ -0,0 +1,56 @@
...
+      class << self
+        def detect_class(payload)
+          function_module_name, function_name =
+            payload.match(INTRINSIC_FUNCTION_REGEX)
+                   .named_captures
+                   .values_at("module", "function")
+
+          begin
+            function_module = Floe::Workflow::IntrinsicFunctions.const_get(function_module_name)
+            function_module.const_get(function_name)
```
Fryguy, line 17: I have one concern here.  I was looking through the [list of Intrinsic Functions](https://states-language.net/#appendix-b) and there is one called `States.Array`.  This would imply we have a `Floe::Workflow::IntrinsicFunctions::States::Array` class, and that `Array` shadowing kind of scares me (regardless of const_get or literal).  I'm wondering if instead of a full namespacing, we instead just smash the module/function together so you'd get something more like: `Floe::Workflow::IntrinsicFunctions::StatesArray`.  Thoughts?

agrare: Hm we could have `States` be a class with methods `def StringToJson` and `def Array` rather than having classes for each intrinsic function.  I like having these separated because if/when we have our own `ManageIQ` intrinsic functions they will be nicely separated.

Fryguy: maybe?  methods named with capitals will surely get confusing.  If they are class methods then, technicall `::` will call those too, which works, but would still be weird.  Reminds me of the Vmdb::BUILD method we have.

kbrock: I like going the `send` route, aka: `def StringToJson` (or `def intrinsicStringToJson` if we need to start with a lowercase). Alternatively a `case` statement may not be the most ruby thing, but that tends to be the defacto implementation for this type of construct across all programming languages (both target language and complier/interpreter language).

All intrinsics are in the same namespace: `States.*`. I think hardcoding this rather than a generic "look for a period" works best, and we can hardcode any additional namespaces in the future. Lets also avoid regular expressions and use `indexOf(")")` kind of stuff. It is way too hard to deal with properly nesting parens and regular expressions. Though this implementation may be ok for a first pass.

PR #85, opened 2023-08-09
Move state from workflow into context

kbrock:
References:
- Extracted from #67
- built upon (to remove codeclimate / failing build) #89

Goal
====

Move the runtime variables from `Workflow` to `Context`. So context will hold all the stateful data. No more methods in workflow that hold some context state.

Tangible goal:

- We can test the various `State` implementations without a `Workflow` instance.
- `State#run` will fully change the context and 

Additions
=======

- Added `NextState`, `Error`, and `Cause` to `Context`. Now errors are exposed.
- Added `context.execution["EndTime"]` and use that for determining `end?` (couldn't find the exact value for that variable)

Changes
=======

- Move hardcoded `State#status` methods to dynamic `Context#status`.
- Definition of `status "pending"` changed a little: it asks "Have I started" instead of "Is my state the starting state".
- Definition of `status "success"` changed a little: it asks "Have I **finished** without error" rather than looking at (`State#next` / `Context#next_state`).

Introduce Facade
=========

I introduced facade `quick_step`. This simplifies the caller in manageiq-provider-workflow's [workflow_instance.rb]:

```ruby
context = Workflow.quick_step(name, payload, context, creds)
update!(:context => context.to_h, :status => context.status, :output => context.output)
```

It is no secret from #67 , that I would like to pass the `context` into `step!` and feel that we could introduce a more sophisticated `step` method - which we can discuss elsewhere.

[workflow_instance.rb]: https://github.com/ManageIQ/manageiq-providers-workflows/blob/3a0900d3435875abd2d8dc9b2e29dcb57d374ecc/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb#L71-L74

### DIFF
--- a/lib/floe/workflow/context.rb
+++ b/lib/floe/workflow/context.rb
@@ -21,8 +21,36 @@ module Floe
         @context["Execution"]
       end
 
+      def started?
+        execution.key?("StartTime")
+      end
+
+      def running?
+        started? && !ended?
+      end
+
+      def ended?
+        execution.key?("EndTime")
+      end
+
       def state
         @context["State"]
+      end
+
+      def output
+        state["Output"]
+      end
+
+      def status
+        if !started?
+          "pending"
+        elsif running?
+          "running"
+        elsif state["Error"]
+          "failure"
+        else
+          "success"
+        end
       end
 
       def state=(val)
--- a/lib/floe/workflow/context.rb
+++ b/lib/floe/workflow/context.rb
@@ -37,8 +37,20 @@ module Floe
         @context["State"]
       end
 
+      def input
+        state["Input"]
+      end
+
       def output
         state["Output"]
+      end
+
+      def state_name
+        state["Name"]
+      end
+
+      def next_state
+        state["NextState"]
       end
 
       def status
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -14,7 +14,7 @@ module Floe
       end
     end
 
-    attr_reader :context, :credentials, :output, :payload, :states, :states_by_name, :current_state, :status
+    attr_reader :context, :credentials, :output, :payload, :states, :states_by_name, :current_state, :status, :start_at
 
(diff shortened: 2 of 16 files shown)


lib/floe/workflow/context.rb
```diff
@@ -21,10 +21,46 @@ def execution
         @context["Execution"]
       end
 
+      def start?
+        execution.key?("StartTime")
+      end
+
+      def end?
+        execution.key?("EndTime")
+      end
```
agrare, line 30: Would `started?` and `running?` make more sense?  `context.running?` I think is a little more obvious than `context.end?`

lib/floe/workflow/context.rb
```diff
@@ -21,10 +21,46 @@ def execution
...
+
+      def status
+        if !start?
+          "pending"
+        elsif !end?
+          "running"
+        elsif state["Error"]
+          "failure"
+        else
+          "success"
```
agrare, line 60: Hm if one step has finished (and thus has set execution["EndTime"] for _that_ state) will this report that the whole workflow is status "success" ?

Should we be clearing the execution Start and End times after the state has run?

agrare: Ah nevermind execution["EndTime"] is only if there is no next state

lib/floe/workflow/context.rb
kbrock: I changed `started?` and `running?` - but left `end?` alone since It is part of the external interface.

Once we change `workflow_instance#run` to `context.running?`, then we can probably drop `context#end?`

agrare:
Overall I like this a lot, helped me with the async runner so I based my branch off of this one

kbrock:
update:
- rebased
- renamed `Context#start?`, `Context#end?` to `started?`, `ended?` and added `running?`
- removed `quick_run`
- added a few specs around `ended?`

agrare (approved):

PR #88, opened 2023-08-10
Fix State Duration units

agrare:
CLOCK_MONOTONIC returns a float with seconds.nanoseconds so just `tick-tock` will result in seconds (float) duration


Fryguy:
🤦 I wrote the PR and then never pushed it up. 🤣   Thanks for getting this one.

PR #89, opened 2023-08-11
fix complexity of parse_payload_hash

kbrock:
This is a minor change that didn't improve the code.
Since it quiets the linter, and doesn't detract too much from the code, it is good enough for me.

fixes code climate issue introduced by #83

### DIFF
--- a/lib/floe/workflow/payload_template.rb
+++ b/lib/floe/workflow/payload_template.rb
@@ -32,9 +32,7 @@ module Floe
       def parse_payload_hash(value)
         value.to_h do |key, val|
           if key.end_with?(".$")
-            if value.key?(key.chomp(".$"))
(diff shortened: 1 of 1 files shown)


lib/floe/workflow/payload_template.rb
```diff
@@ -32,9 +32,7 @@ def parse_payload_array(value)
       def parse_payload_hash(value)
         value.to_h do |key, val|
           if key.end_with?(".$")
-            if value.key?(key.chomp(".$"))
-              raise Floe::InvalidWorkflowError, "both #{key} and #{key.chomp(".$")} present"
-            end
+            check_key_conflicts(value, key)
```
Fryguy, line 35: super-extreme-minor, but my brain wants to see `key, value` in that order, and seeing it backwards is weird - can we flip these?

agrare (approved):

PR #92, opened 2023-09-07
Add a nonblocking workflow step function

agrare:
Add a non-blocking way to step through a workflow.  If a state doesn't won't block then it behaves exactly the same (e.g. choice, pass, etc...).  If a state would block then `step_nonblock` will return `Errno::EAGAIN` allowing the caller to break out and move on to other work.  Calling `step_nonblock` again once the task is complete will finish the rest of the work and cleanup from the run.

Future TODO:
- [ ] Add an event-driven `Workflow#wait` method (e.g. using k8s watches) https://github.com/ManageIQ/floe/pull/95
- [ ] Add a class-level `Workflow.wait` that can wait for multiple workflows

Depends on:
- [x] https://github.com/ManageIQ/floe/pull/52

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,19 +81,14 @@ would block will return `Errno::EAGAIN`.
 require 'floe'
 
 workflow = Floe::Workflow.load("workflow.asl")
-workflow.run_async!
 
-until workflow.end?
-  loop while workflow.step_nonblock == 0 # Step through the workflow while it would not block
-  break
-end
+# Step through the workflow while it would not block
+workflow.run_nonblock
 
 # Go off and do some other task
 
-until workflow.end?
-  loop while workflow.step_nonblock == 0 # Continue stepping until the workflow is finished
-  break
-end
+# Continue stepping until the workflow is finished
+workflow.run_nonblock
 ```
 
 You can also use the `Floe::Workflow.wait` class method to wait on multiple workflows
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ bundle exec ruby exe/floe --workflow my-workflow.asl --credentials='{"roleArn": 
 ```ruby
 require 'floe'
 
-workflow = Floe::Workflow.load(File.read("workflow.asl"))
+workflow = Floe::Workflow.load("workflow.asl")
 workflow.run!
 ```
 
@@ -68,8 +68,54 @@ Floe::Workflow::Runner.docker_runner = Floe::Workflow::Runner::Podman.new
 # Or
 Floe::Workflow::Runner.docker_runner = Floe::Workflow::Runner::Kubernetes.new("namespace" => "default", "server" => "https://k8s.example.com:6443", "token" => "my-token")
 
-workflow = Floe::Workflow.load(File.read("workflow.asl"))
+workflow = Floe::Workflow.load("workflow.asl")
 workflow.run!
+```
+
+### Non-Blocking Workflow Execution
+
+It is also possible to step through a workflow without blocking, and any state which
+would block will return `Errno::EAGAIN`.
+
+```ruby
+require 'floe'
+
+workflow = Floe::Workflow.load("workflow.asl")
+workflow.run_async!
+
+until workflow.end?
+  loop while workflow.step_nonblock == 0 # Step through the workflow while it would not block
+  break
+end
+
+# Go off and do some other task
+
+until workflow.end?
+  loop while workflow.step_nonblock == 0 # Continue stepping until the workflow is finished
+  break
+end
+```
+
+You can also use the `Floe::Workflow.wait` class method to wait on multiple workflows
+and return all that are ready to be stepped through.
+
+```ruby
+require 'floe'
+
+workflow1 = Floe::Workflow.load("workflow1.asl")
+workflow2 = Floe::Workflow.load("workflow2.asl")
+
+running_workflows = [workflow1, workflow2]
+until running_workflows.empty?
+  # Wait for any of the running workflows to be ready (up to the timeout)
+  ready_workflows = Floe::Workflow.wait(running_workflows)
+  # Step through the ready workflows until they would block
+  ready_workflows.each do |workflow|
+    loop while workflow.step_nonblock == 0
+  end
+  # Remove any finished workflows from the list of running_workflows
+  running_workflows.reject!(&:end?)
+end
 ```
 
 ### Docker Runner Options
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -126,12 +126,9 @@ module Floe
     end
 
     def step_nonblock_finish
-      current_state.finish_async
-      context.state["FinishedTime"] = Time.now.utc
-      context.state["Duration"]     = context.state["FinishedTime"] - context.state["EnteredTime"]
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -68,10 +68,56 @@ Floe::Workflow::Runner.docker_runner = Floe::Workflow::Runner::Podman.new
...
+```ruby
+require 'floe'
+
+workflow = Floe::Workflow.load("workflow.asl")
+workflow.run_async!
+
+until workflow.end?
+  loop while workflow.step_nonblock == 0 # Step through the workflow while it would not block
+  break
+end
```
kbrock, line 89: Maybe

```ruby
class Workflow
  # run_nonblock
  def run_until_block
    loop while step_nonblock == 0 unless end?
  end
end
```

I'm thinking the `until` and the `break` together are just an `if`?

Regardless of what the code looks like, introducing a method to make documentation simpler seems like a good plan.

lib/floe/workflow.rb
```diff
@@ -34,43 +51,45 @@ def initialize(payload, context = nil, credentials = {})
...
 
-    def step
-      context.execution["StartTime"] ||= Time.now.utc
-
-      context.state["Guid"]    = SecureRandom.uuid
-      context.state["Input"] ||= context.execution["Input"].dup
+    def run!
+      until end?
+        step
+      end
```
kbrock, line 57: `step` is non-blocking.

So I think this could spin a lot of cpu.

Probably for another spike PR:
Does it make sense to return the number of seconds to check back in?
If it encoutered a wait/retry, then it could return the value
If it is a task (and unknown), then it just returns 1?

I say this is for another PR because that would potentially make it a more involved/confusing interface.

lib/floe/workflow.rb
```diff
@@ -84,5 +103,40 @@ def output
...
+
+    private
+
+    def step_nonblock_submit
+      raise "State is already running" if current_state.started?
+
+      start_time = Time.now.utc
+
+      context.execution["StartTime"] ||= start_time
+      context.state["Input"]         ||= context.execution["Input"].dup
```
kbrock, line 119: Can we always set input and State (from start_at) at the same time?
I liked the idea of guid at that same time.

we agree execution and state starting time/entered time need to be carefully thought through.

lib/floe/workflow/state.rb
```diff
@@ -29,9 +29,32 @@ def initialize(workflow, name, payload)
         @comment  = payload["Comment"]
       end
 
+      def run!(input)
+        run_async!(input)
+        sleep(1) while running?
```
kbrock, line 34: almost feel like dropping `State#run!` and moving that blocking knowledge into `Workflow` may make sense. thought this concept is pretty simple

lib/floe/workflow/state.rb
```diff
@@ -29,9 +29,32 @@ def initialize(workflow, name, payload)
...
+        finish
```
kbrock, line 35: Fun when we arrive at the same concept (albeit in separate threads)

lib/floe/workflow/context.rb
```diff
@@ -45,6 +45,10 @@ def output
         state["Output"]
       end
 
+      def output=(val)
+        state["Output"] = val
+      end
+
```
kbrock, line 51: I admit that code climate complaining about adding too many methods to `Context` shaped my design. So I had a single method that sets the `output`, `next_state`, `error`, and `cause`.

Nice thing about these helper methods is we can move `output` out of `"State"` when we want to remove some of these non-spec values from the hash.

kbrock (commented):
just a few thoughts that were probably already mentioned in 94 or in chat

spec/workflow_spec.rb
```diff
@@ -75,38 +73,156 @@
...
+        expect(workflow.end?).to   eq(false)
+      end
+    end
+
+    context "with a running state" do
+      it "returns Try again" do
+        workflow = make_workflow(input, {"FirstState" => {"Type" => "Wait", "Seconds" => 10, "End" => true}})
+        expect(workflow.step_nonblock).to eq(Errno::EAGAIN)
+      end
+    end
```
kbrock, line 133: what does this buy us?

If it said it wanted 10 seconds.
And you called back in 8 seconds, and it wasn't ready, why not just have it return basically saying "call back in 2 seconds or something"?

lib/floe/workflow/states/fail.rb
```diff
@@ -13,8 +13,13 @@ def initialize(workflow, name, payload)
           @error = payload["Error"]
         end
 
-        def run!(input)
-          [nil, input]
+        def run_async!(input)
+          context.next_state = nil
+          context.output     = input
```
kbrock, line 18: need to set cause and error

lib/floe/workflow.rb
kbrock: 1. I prefer if we set input to execution input at the same place we set the initial state
2. I would prefer if we did not set the first state in the workflow#initialize, but that is all wrapped up in passing context into `step` so we can worry about that later.

lib/floe/workflow/state.rb
agrare: Yeah I think you're right not sure what this buys us since no one should be hitting the state directly.  Only callers right now are in state specs.

README.md
agrare: Okay how about instead of `Workflow#run_async!` we have `Workflow#run_nonblock` which behaves similarly to `step_nonblock` but incorporates your loop.

spec/workflow_spec.rb
agrare: I wouldn't mind a method that allowed a caller to ask when to check a state again in cases where we know the ready time (wait, task+catch+retry).  I'd like to tackle this in a future enhancement, I was planning on doing something like this for the wait method that is waiting on many states and set the wakeup to the first Wait completion this would be a good way to get that time generically.

lib/floe/workflow/states/fail.rb
agrare: I had that in finish, though not sure why and it works just as well here and I can drop the `finish` override

lib/floe/workflow.rb
```diff
@@ -34,43 +51,45 @@ def initialize(payload, context = nil, credentials = {})
...
+      return Errno::EPERM if end?
 
-      context.state["FinishedTime"] = Time.now.utc
-      context.state["Duration"]     = tock - tick
-      context.state["Output"]       = output
-      context.state["NextState"]    = next_state
-      context.state["Error"]        = current_state.error if current_state.respond_to?(:error)
-      context.state["Cause"]        = current_state.cause if current_state.respond_to?(:cause)
-      context.execution["EndTime"]  = Time.now.utc if next_state.nil?
+      step_nonblock_submit unless current_state.started?
```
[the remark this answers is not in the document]
agrare: I really don't like having Context start/end the states when we have a State class, if should be that class' responsibility even if things end up being set in Context IMO.

I know it isn't exactly what you're looking for but I added `State#start` to replace `State#run_async!` (async on all states didn't make sense with what I was saying anyway and was a holdover from really early implementation).  I'm thinking when we ask a state to start that is where we set all of these `context.state` initial values.

PR #94, opened 2023-09-08
Context state run

kbrock:
Extracted from https://github.com/ManageIQ/floe/pull/67

- `Workflow#run` handles contexts not yet run, not completed, completed/waiting to run next state, and completed/finished
  - no longer modify `context` in `Workflow#initialize`
  - no longer auto advance to the next state
- Context helper methods for setting at beginning and end.
- `State#run` changes context. (currently ignoring `_input` parameter)

I did not remove `context`/`workflow` from `State#initialize`. This left the input for `run()` a little in limbo.

### DIFF
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -28,17 +28,17 @@ module Floe
 
       @states         = payload["States"].to_a.map { |name, state| State.build!(self, name, state) }
       @states_by_name = @states.each_with_object({}) { |state, result| result[state.name] = state }
-
-      context.state["Name"] ||= start_at
     rescue JSON::ParserError => err
       raise Floe::InvalidWorkflowError, err.message
     end
 
     def step
-      context.execution["StartTime"] ||= Time.now.utc
-
-      context.state["Guid"]    = SecureRandom.uuid
-      context.state["Input"] ||= context.execution["Input"].dup
+      if context.next_state
+        context.state = {"Name" => context.next_state, "Input" => context.output, "Guid" => SecureRandom.uuid}
+      elsif !context.state_name
+        context.execution["StartTime"] = Time.now.utc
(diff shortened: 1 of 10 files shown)


kbrock:
update:
- fixed spacing cops

kbrock:
update:
- fixed whitespace
- merged `start_next_state` and `start_workspace` to remove 1 method (to make code climate more comfortable)

kbrock:
update:
- In specs: fixed indentation (rubocop)
- In specs: fixed not passing ctx into a few workflow.load calls

kbrock:
This is starting to overlap https://github.com/ManageIQ/floe/pull/92 a bit

lib/floe/workflow.rb
```diff
@@ -51,11 +50,11 @@ def step
...
-      context.state["Output"]       = output
-      context.state["NextState"]    = next_state
-      context.state["Error"]        = current_state.error if current_state.respond_to?(:error)
-      context.state["Cause"]        = current_state.cause if current_state.respond_to?(:cause)
-      context.execution["EndTime"]  = Time.now.utc if next_state.nil?
+      if current_state.respond_to?(:error)
+        context.end_state!(output, :error => current_state.error, :cause => current_state.cause)
+      else
+        context.end_state!(output, next_state)
+      end
```
agrare: What if we move this into `State` rather than `Context`?  I think the fact that we're checking `State#respond_to?(:error)` is an indication that this is in the wrong place.

When I think of setting initial `context.state` values it sounds like something that is done by `State#run*` or `State#before_start`

kbrock: yea. I want to move there
There reason why I was ok with `respond_to?` is because we are moving that call into `State` in the short term. (either your PR or mine)

kbrock:
most of these concepts are in other PRs

PR #95, opened 2023-09-12
Add a watch method to Workflow::Runner for event driven updates

agrare:
Add a method for getting update driven notifications about floe events

Get all updates indefinitely:

```ruby
Floe::Workflow::Runner.docker_runner.wait { |runner_context| puts runner_context }
```

Get only create/update events for 30 seconds then break out:

```ruby
Floe::Workflow::Runner.wait(:timeout => 30, :events => %i[create update]) { |runner_context| puts runner_context }
```

TODO
- [x] Kubernetes watches
- [x] Docker events
- [x] Podman events
- [x] Return a compatible runner context from the wait
- [x] Add a way to terminate the wait loop (accept a timeout option)
- [x] Update `Floe::Workflow.wait` to use `Floe::Workflow::Runner#wait` events
- [x] Update exe/floe to use `Floe::Workflow.wait`

### DIFF
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -20,29 +20,67 @@ module Floe
         workflows = [workflows] if workflows.kind_of?(self)
         logger.info("checking #{workflows.count} workflows...")
 
-        start = Time.now.utc
-        ready = []
+        run_until   = Time.now.utc + timeout if timeout.to_i > 0
+        ready       = []
+        queue       = Queue.new
+        wait_thread = Thread.new do
+          loop do
+            Runner.for_resource("docker").wait do |event, runner_context|
+              queue.push([event, runner_context])
+            end
+          end
+        end
 
         loop do
           ready = workflows.select(&:step_nonblock_ready?)
-          break if timeout&.zero?
-          break if timeout && Time.now.utc - start > timeout
+          break if block.nil? && !ready.empty?
 
-          unless ready.empty?
-            if block
-              ready.each(&block)
-            else
-              break
+          ready.each(&block)
+
+          # Break if all workflows are completed or we've exceeded the
+          # requested timeout
+          break if workflows.all?(&:end?)
+          break if timeout && (timeout.zero? || Time.now.utc > run_until)
+
+          # Find the earliest time that we should wakeup if no container events
+          # are caught, either a workflow in a Wait or Retry state or we've
+          # exceeded the requested timeout
+          wait_until = workflows.map(&:wait_until)
+                                .unshift(run_until)
+                                .compact
+                                .min
+
+          # If a workflow is in a waiting state wakeup the main thread when
+          # it will be done sleeping
+          if wait_until
+            sleep_thread = Thread.new do
+              sleep wait_until - Time.now.utc
+              queue.push(nil)
             end
           end
 
-          break if workflows.all?(&:end?)
+          loop do
+            # Block until an event is raised
+            event, runner_context = queue.pop
+            break if event.nil?
 
-          sleep(1)
+            # If the event is for one of our workflows set the updated runner_context
+            workflows.each do |workflow|
+              next unless workflow.context.state.dig("RunnerContext", "container_ref") == runner_context["container_ref"]
+
+              workflow.context.state["RunnerContext"] = runner_context
+            end
+
+            break if queue.empty?
+          end
+        ensure
+          sleep_thread&.kill
         end
 
         logger.info("checking #{workflows.count} workflows...Complete - #{ready.count} ready")
         ready
+      ensure
+        wait_thread&.kill
       end
     end
 
@@ -92,6 +130,14 @@ module Floe
 
     def step_nonblock_ready?
       current_state.ready?
+    end
+
+    def waiting?
+      current_state.waiting?
+    end
+
+    def wait_until
+      current_state.wait_until
     end
 
     def status
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -17,6 +17,7 @@ module Floe
       end
 
       def wait(workflows, timeout: nil)
+        workflows = [workflows] if workflows.kind_of?(self)
         logger.info("checking #{workflows.count} workflows...")
 
         start = Time.now.utc
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -16,7 +16,7 @@ module Floe
         new(payload, context, credentials, name)
       end
 
-      def wait(workflows, timeout: nil)
+      def wait(workflows, timeout: nil, &block)
         workflows = [workflows] if workflows.kind_of?(self)
         logger.info("checking #{workflows.count} workflows...")
 
@@ -25,7 +25,18 @@ module Floe
 
         loop do
           ready = workflows.select(&:step_nonblock_ready?)
-          break if timeout.zero? || Time.now.utc - start > timeout || !ready.empty?
+          break if timeout&.zero?
+          break if timeout && Time.now.utc - start > timeout
+
+          unless ready.empty?
+            if block
+              ready.each(&block)
+            else
+              break
+            end
+          end
+
+          break if workflows.all?(&:end?)
 
           sleep(1)
         end
--- a/lib/floe/workflow/runner/docker.rb
+++ b/lib/floe/workflow/runner/docker.rb
@@ -76,7 +76,7 @@ module Floe
               ref, event = parse_notice(notice)
               next unless events.include?(event)
 
-              state = inspect_container(ref).first&.dig("State")
+              state = inspect_container(ref)&.dig("State")
 
               runner_context = {"container_ref" => ref, "container_state" => state}
 
@@ -105,7 +105,7 @@ module Floe
         def status!(runner_context)
           return if runner_context.key?("Error")
 
-          runner_context["container_state"] = inspect_container(runner_context["container_ref"]).first&.dig("State")
+          runner_context["container_state"] = inspect_container(runner_context["container_ref"])&.dig("State")
         end
 
         def running?(runner_context)
@@ -174,7 +174,9 @@ module Floe
         end
 
         def inspect_container(container_id)
-          JSON.parse(docker!("inspect", container_id).output)
+          JSON.parse(docker!("inspect", container_id).output).first
+        rescue
+          nil
         end
 
         def delete_container(container_id)
--- a/lib/floe/workflow/runner/docker.rb
+++ b/lib/floe/workflow/runner/docker.rb
@@ -45,6 +45,9 @@ module Floe
           delete_secret(secrets_file)    if secrets_file
         end
 
+        def wait(timeout: nil, events: %i[create update delete])
+        end
+
         def status!(runner_context)
           return if runner_context.key?("Error")
 
--- a/lib/floe/workflow/runner/docker.rb
+++ b/lib/floe/workflow/runner/docker.rb
@@ -76,7 +76,11 @@ module Floe
               ref, event = parse_notice(notice)
               next unless events.include?(event)
 
-              notices << [ref, event]
+              state = inspect_container(ref).first&.dig("State")
+
+              runner_context = {"container_ref" => ref, "container_state" => state}
+
+              notices << runner_context
             end
 
             # If we're given a block yield the events otherwise return them
@@ -147,6 +151,7 @@ module Floe
         def wait_params(until_timestamp)
           params = ["events", [:format, "{{json .}}"], [:filter, "type=container"], [:since, Time.now.utc.to_i]]
           params << [:until, until_timestamp.to_i] if until_timestamp
+          params
         end
 
         def parse_notice(notice)
--- a/lib/floe/workflow/runner/docker.rb
+++ b/lib/floe/workflow/runner/docker.rb
@@ -78,9 +78,7 @@ module Floe
 
               state ||= inspect_container(ref)&.dig("State")
 
-              runner_context = {"container_ref" => ref, "container_state" => state}
-
-              notices << runner_context
+              notices << {"container_ref" => ref, "container_state" => state}
             end
 
             # If we're given a block yield the events otherwise return them
--- a/lib/floe/workflow/runner/docker.rb
+++ b/lib/floe/workflow/runner/docker.rb
@@ -76,7 +76,7 @@ module Floe
               event, runner_context = parse_notice(notice)
               next unless events.include?(event)
 
-              notices << runner_context
+              notices << [event, runner_context]
             end
 
             # If we're given a block yield the events otherwise return them
--- a/lib/floe/workflow/runner/docker.rb
+++ b/lib/floe/workflow/runner/docker.rb
@@ -164,7 +164,7 @@ module Floe
             :create
           when "start"
             :update
-          when "die"
+          when "die", "destroy"
             :delete
           else
             :unkonwn
--- a/lib/floe/workflow/runner/docker.rb
+++ b/lib/floe/workflow/runner/docker.rb
@@ -73,10 +73,10 @@ module Floe
               # always `ready?`
               break if notice.nil?
 
-              ref, event = parse_notice(notice)
+              ref, event, state = parse_notice(notice)
               next unless events.include?(event)
 
-              state = inspect_container(ref)&.dig("State")
+              state ||= inspect_container(ref)&.dig("State")
 
               runner_context = {"container_ref" => ref, "container_state" => state}
 
--- a/lib/floe/workflow/runner/docker.rb
+++ b/lib/floe/workflow/runner/docker.rb
@@ -10,6 +10,7 @@ module Floe
 
         def initialize(options = {})
           require "awesome_spawn"
+          require "io/wait"
           require "tempfile"
 
           super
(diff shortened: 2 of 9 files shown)


lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -78,6 +78,15 @@ def output(runner_context)
           runner_context["output"] = output
         end
 
+        def wait(contexts, timeout: 5)
+          contexts = [contexts] unless contexts.kind_of?(Array)
+          refs     = contexts.map { |ctx| ctx["container_ref"] }
```
kbrock, line 83: Does `Array.wrap(contexts) meet our needs

```suggestion
          refs     = Array.wrap(contexts).map { |ctx| ctx["container_ref"] }
```

lib/floe/workflow/state.rb
```diff
@@ -97,9 +97,15 @@ def finished?
         context.state.key?("FinishedTime")
       end
 
+      def wait_until
+        return nil unless context.state.key?("WaitUntil")
+
+        Time.parse(context.state["WaitUntil"])
```
kbrock, line 103:
```suggestion
        Time.parse(context.state["WaitUntil"]) if context.state.key?("WaitUntil")
```

lib/floe/workflow/runner/kubernetes.rb
agrare: :+1: No activesupport so no `Array.wrap()` but we can use `Array(contexts).map`

kbrock:
I'm getting distracted by `Rename State#run_wait to just #wait`
Can you pull that out to a separate PR?
It is a slam dunk.

Lol. Code climate is asking for you to add back `please_hold()`. I'm ok with it gone.
I had tried hard to find a way to get the complexity there down. But 4 variables that are basically the same thing just plain make it complicated for the machine (and simple for you and me)

agrare:
> Lol. Code climate is asking for you to add back please_hold(). I'm ok with it gone.

Ahh that makes so much more sense, was wondering why we pulled that out into a separate method when it was only called in one place.

agrare:
> I'm getting distracted by Rename State#run_wait to just #wait
> Can you pull that out to a separate PR?

:+1: yes great idea

lib/floe/workflow/runner/docker.rb
```diff
@@ -60,6 +63,62 @@ def output(runner_context)
           runner_context["output"] = output
         end
 
+        def wait(contexts = [], timeout: nil, since: Time.now.utc, events: [], &block)
+          refs              = Array(contexts).map { |ctx| ctx["container_ref"] }
+          container_filters = refs.map   { |container_id| [:filter, "container=#{container_id}"] } + [[:filter, "type=container"]]
```
Fryguy, line 68: Minor perf improvement:

```suggestion
          container_filters = refs.map   { |container_id| [:filter, "container=#{container_id}"] } << [:filter, "type=container"]
```

lib/floe/workflow/runner/docker.rb
```diff
@@ -60,6 +63,62 @@ def output(runner_context)
...
+
+          until_timestamp = Time.now.utc + timeout if timeout
+
+          r, w = IO.pipe
+
+          params = ["events", [:format, "{{json .}}"]]
+          params += event_filters        if event_filters.any?
+          params += container_filters    if container_filters.any?
+          params += [:since, since.to_i] if since
+          params += [:until, until_timestamp.to_i] if until_timestamp
```
Fryguy, line 79: Similarly here, prefer << (or .concat) over += where we can

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -92,6 +92,21 @@ def output(runner_context)
...
+            ref = notice.object.metadata.name
+            next unless refs.empty? || refs.include?(ref)
+
+            if block
+              yield ref
+            else
+              return ref
+            end
+          end
+        end
```
Fryguy, line 108: Since we're not actually using the block here, I think it's a little more performant to not capture it into a variable.

```suggestion
        def wait(contexts = [], timeout: nil, since: Time.now.utc, events: [])
          refs = Array(contexts).map { |ctx| ctx["container_ref"] }

          kubeclient.watch_pods(:namespace => namespace) do |notice|
            ref = notice.object.metadata.name
            next unless refs.empty? || refs.include?(ref)

            if block_given?
              yield ref
            else
              return ref
            end
          end
        end
```

Fryguy: Unless you're trying to keep it consistent with the other runner, which is fine too

lib/floe/workflow/runner/docker.rb
```diff
@@ -90,15 +149,28 @@ def delete_container(container_id)
           nil
         end
 
+        def delete_secret(secrets_file)
+          return unless File.exist?(secrets_file)
+
+          File.unlink(secrets_file)
```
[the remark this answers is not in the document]
agrare: Because unlink is the real method that it calls and that's what I'm used to :shrug:

lib/floe/workflow/runner/docker.rb
```diff
@@ -60,6 +63,62 @@ def output(runner_context)
...
+          container_filters = refs.map   { |container_id| [:filter, "container=#{container_id}"] } << [:filter, "type=container"]
+          event_filters     = events.map { |event| [:filter, "event=#{event}"] }
+
+          until_timestamp = Time.now.utc + timeout if timeout
+
+          r, w = IO.pipe
+
+          params = ["events", [:format, "{{json .}}"]]
+          params += event_filters                  if event_filters.any?
+          params += container_filters              if container_filters.any?
```
kbrock, line 77: this will always have a value

agrare: true, had this before I added the type=container filter

kbrock: pulling out a new method `build_params` may trim down `wait` enough to make code climate happier?

It would contain  (refs, container_filter, event_filter, params)

lib/floe/workflow/runner/docker.rb
```diff
@@ -45,10 +46,66 @@ def cleanup(runner_context)
...
+          )
+
+          w.close
+
+          loop do
+            readable_timeout = until_timestamp - Time.now.utc if until_timestamp
+
+            # Wait for our end of the pipe to be readable and if it didn't timeout
+            # get the events from stdout
+            next if r.wait_readable(readable_timeout).nil?
```
Fryguy, line 65: Just confirming, but readable_timeout here is in seconds (Because Time#- returns the difference in seconds).  Is that the expected unit for wait_readable?

lib/floe/workflow/runner/docker.rb
```diff
@@ -91,8 +148,35 @@ def run_container_params(image, env, secrets_file)
...
         end
 
+        def wait_params(until_timestamp)
+          params = ["events", [:format, "{{json .}}"], [:filter, "type=container"], [:since, Time.now.utc.to_i]]
+          params << [:until, until_timestamp.to_i] if until_timestamp
+          params
+        end
+
+        def parse_notice(notice)
+          notice = JSON.parse(notice)
```
Fryguy, line 158: Do we have any handling for if this doesn't parse? (I realize that's unlikely but I could see things where a partial unbuffered write is terminated mid-write and you don't get a full JSON line.

lib/floe/workflow/runner/docker.rb
```diff
@@ -45,10 +46,66 @@ def cleanup(runner_context)
...
+              notice = r.gets
+
+              # If the process has exited `r.gets` returns `nil` and the pipe is
+              # always `ready?`
+              break if notice.nil?
+
+              ref, event, state = parse_notice(notice)
+              next unless events.include?(event)
+
+              state ||= inspect_container(ref)&.dig("State")
```
Fryguy, line 79: Can you explain/comment what case would cause state to be nil and that we'd need to do the follow-up docker inspect (which I expect is expensive, hence why I'm asking)

Fryguy, line 79: Can you explain/comment what case would cause state to be nil and that we'd need to do the follow-up docker inspect (which I expect is expensive, hence why I'm asking)

lib/floe/workflow/runner/docker.rb
```diff
@@ -45,10 +46,66 @@ def cleanup(runner_context)
...
+
+              runner_context = {"container_ref" => ref, "container_state" => state}
+
+              notices << runner_context
```
Fryguy, line 83: Super-minor, but there's no need for a temp var from what I can see
```suggestion
              notices << {"container_ref" => ref, "container_state" => state}
```

lib/floe/workflow/runner/docker.rb
agrare: On podman we can get the state right from the event, on docker we have to do an explicit inspect

lib/floe/workflow/runner/docker.rb
```diff
@@ -45,10 +46,66 @@ def cleanup(runner_context)
...
+            if block
+              notices.each(&block)
+            else
+              # Terminate the `docker events` process before returning the events
+              Process.kill("TERM", pid) rescue Errno::ESRCH
+              return notices
+            end
+
+            # Check that the `docker events` process is still alive
+            Process.kill(0, pid)
```
Fryguy, line 96: Does this block?  I think I'm confused how we will break out of the loop if the docker events process hasn't exited in time.

lib/floe/workflow/runner/docker.rb
```diff
@@ -45,10 +46,66 @@ def cleanup(runner_context)
...
+          rescue Errno::ESRCH
+            # Break out of the loop if the `docker events` process has exited
+            break
+          end
+        ensure
+          r.close
```
Fryguy, line 102: Does this need a rescue/check in case the pipe is closed prematurely or is that not possible? (I'm thinking if there's an error and the pipe is broken).  Maybe something like the following (or maybe close just handles it already being closed and this all doesn't matter)

```suggestion
          r.close unless r.closed?
```

lib/floe/workflow/runner/docker.rb
Fryguy: Actually, unless I'm reading this incorrectly, state can _never_ be nil because docker_event_status_to_event never returns nil (it returns :unknown in the unknown case.

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -102,6 +102,33 @@ def cleanup(runner_context)
           delete_secret(secret) if secret
         end
 
+        def wait(timeout: nil, events: %i[create update delete])
```
Fryguy, line 105: Out of curiosity was this pattern pulled from one of the providers?  I'm just curious how you knew to handle things like the timeouts happening at interesting times.

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -102,6 +102,33 @@ def cleanup(runner_context)
...
+          watcher = kubeclient.watch_pods(:namespace => namespace)
+
+          timeout_thread = Thread.new { sleep(timeout); watcher.finish } if timeout.to_i > 0
+
+          watcher.each do |notice|
```
Fryguy, line 110: Out of curiosity was this pattern pulled from one of the providers?  I'm just curious how you knew to handle things like the timeouts happening at interesting times.

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -102,6 +102,33 @@ def cleanup(runner_context)
...
+            event = kube_notice_type_to_event(notice.type)
+            next unless events.include?(event)
+
+            pod = notice.object
+            container_ref   = pod.metadata.name
+            container_state = pod.to_h.deep_stringify_keys["status"]
```
Fryguy, line 116: Why the deep_stringify_keys here? Seems like an expensive operation just to get the status, where instead you could do

```suggestion
            pod_h = pod.to_h
            container_state = pod_h.fetch("status", pod_h[:status])
```

Or perhaps just hit the pod object direct with something like `pod.status`

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -102,6 +102,33 @@ def cleanup(runner_context)
...
+
+            runner_context = {"container_ref" => container_ref, "container_state" => container_state}
+
+            if block_given?
+              yield runner_context
+            else
+              timeout_thread&.kill # If we break out before the timeout, kill the timeout thread
+              return runner_context
```
Fryguy, line 124: I noticed that the docker one builds up an array of notices and returns all of them, whereas this one only returns a single context.  That means the return interface is inconsistent (Array of Hash vs just a single Hash).  Is that intentional because of how kube works? Or perhaps this should return `[runner_context]`?

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -102,6 +102,33 @@ def cleanup(runner_context)
           delete_secret(secret) if secret
         end
 
+        def wait(timeout: nil, events: %i[create update delete])
+          watcher = kubeclient.watch_pods(:namespace => namespace)
+
+          timeout_thread = Thread.new { sleep(timeout); watcher.finish } if timeout.to_i > 0
```
Fryguy, line 108: rubocop doesn't like this, and while I like conciseness perhaps we should expand it out (especially if we need more lines later)

```suggestion
          if timeout.to_i > 0
            timeout_thread = Thread.new do
              sleep(timeout)
              watcher.finish
            end
          end
```

lib/floe/workflow/runner/podman.rb
```diff
@@ -55,6 +55,30 @@ def delete_secret(secret_guid)
           nil
         end
 
+        def parse_notice(notice)
+          id, status, exit_code = JSON.parse(notice).values_at("ID", "Status", "ContainerExitCode")
+
+          event   = podman_event_status_to_event(status)
+          running = event != :delete
+
+          container_state = {"Running" => running, "ExitCode" => exit_code.to_i}
```
Fryguy, line 64: What defines the container state format?  That is, they seems different across all 3, and I can't tell if that's important or not (it feels like they should all return a consistent state format)

lib/floe/workflow/runner/docker.rb
agrare: `Process.kill` does not block, if the `docker events` process has exited then `Process.kill` raises `No such process (Errno::ESRCH)` (see below)

lib/floe/workflow/runner/docker.rb
agrare: On docker this will return just `[ref, event]` so yes state will be `nil`.

lib/floe/workflow/runner/docker.rb
agrare: > or maybe close just handles it already being closed and this all doesn't matter

This, if it were somehow closed already `r.close` doesn't raise an exception
```
>> r, w = IO.pipe
=> [#<IO:fd 5>, #<IO:fd 6>]
>> r.close
=> nil
>> r.close
=> nil
>> r.close
=> nil
```

lib/floe/workflow/runner/podman.rb
agrare: It is intended to be "opaque" and only accessed by the runner so it doesn't have to be consistent (the pod inspect output is different between all three, external callers use the `running?` and `success?` interface methods.

Currently we're only checking Running and ExitCode so it could potentially be limited to just on all three that but it was originally the full `docker inspect` output

lib/floe/workflow/runner/kubernetes.rb
agrare: Changed this to return `[]`, did the kubernetes one first then found docker events could return multiple.  They both yield individual items at a time but kubernetes can't return multiple.

lib/floe/workflow/runner/kubernetes.rb
agrare: We have issues with symbol keys and serializing to jsonb and back, that said this could be `pod.to_h[:status].deep_stringify_keys` instead

lib/floe/workflow/runner/docker.rb
agrare: I found a way to get exitCode from docker events so I'm going to drop this

Fryguy: > On docker this will return just `[ref, event]` so yes state will be `nil`.

ahhh I mised the 3 params return vs 2.  Thanks!

Fryguy:
Overall LGTM - only questions I have left are about possible JSON parsing failures as we are parsing the notices line by line.

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -102,6 +102,64 @@ def cleanup(runner_context)
...
+            watcher = kubeclient.watch_pods(:namespace => namespace)
+
+            retry_connection = true
+
+            if timeout.to_i > 0
+              timeout_thread = Thread.new do
+                sleep(timeout)
+                watcher.finish
+              end
+            end
```
kbrock, line 118: not thrilled with this.
But it is a very common pattern (i.e.: feels like go code)

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -102,6 +102,64 @@ def cleanup(runner_context)
...
+                logger.warn("Received [#{code} #{reason}], [#{message}]")
+                break
+              end
+
+              event = kube_notice_type_to_event(notice.type)
+              next unless events.include?(event)
+
+              pod = notice.object
+              container_ref   = pod.metadata.name
+              container_state = pod.to_h[:status].deep_stringify_keys
```
kbrock, line 135: something like the parse_notice code would be nice here.
Think that would make CodeClimate much happier.

This is probably good enough for now and can be done later

kbrock (approved):

lib/floe/workflow/runner/kubernetes.rb
agrare: Yeah it is my preference for any method call that could block indefinitely to take a timeout option, otherwise this is what you have to do.

`kubeclient.watch_pods(:timeout => 30.seconds)` would fix this :)

agrare:
@kbrock @Fryguy this should be ready to go now

lib/floe/workflow.rb
```diff
@@ -16,21 +16,71 @@ def load(path_or_io, context = nil, credentials = {}, name = nil)
...
-      def wait(workflows, timeout: nil)
+      def wait(workflows, timeout: nil, &block)
+        workflows = [workflows] if workflows.kind_of?(self)
         logger.info("checking #{workflows.count} workflows...")
 
-        start = Time.now.utc
-        ready = []
+        run_until   = Time.now.utc + timeout if timeout.to_i > 0
+        ready       = []
+        queue       = Queue.new
```
Fryguy, line 25: Is there a concurrency concern here?  i.e. should we be using something from concurrent-ruby, or is Queue thread safe enough?

agrare: Is `Queue` not thread-safe?  We use it in ManageIQ core across threads in a few places

Fryguy (approved):
Super-high-level I'm good with this PR (have one non-blocking question above).

In the future I wonder if it makes more sense to use something like concurrent-ruby Promises or some other abstraction.  I do worry that there are hidden synchronization bugs here cause threads are hard.

lib/floe/workflow.rb
agrare: The docs for `Queue` show examples using multiple threads

agrare:
> In the future I wonder if it makes more sense to use something like concurrent-ruby Promises or some other abstraction. I do worry that there are hidden synchronization bugs here cause threads are hard.

So one of my biggest issues with `Queue` is there is no "timed-dequeue" so I had to add an extra thread that just did sleep+queue.put(nil).  I was able to get around this by adding a Concurrent::Semaphore that has try_acquire with a timeout parameter but it started looking gross.

lib/floe/workflow.rb
Fryguy: I can never remember hence my question - Some thing in Ruby are thread safe, others are partially and others are not.  Otherwise there wouldn't be a Concurrent::Hash :(

PR #97, opened 2023-09-20
Move workflow#step into State#run

kbrock:
## Overview

Move `Workflow#step_nonblock` to `State#run_nonblock` and related variables.
`State` now owns the complete run cycle.

## Details

- This is mostly moving code and not changing anything.
- `State#run` (blocking) is the only change, and you will recognize that code as the standard `wait until run` style.
- Step through commits. none of them seem to do anything

## Side effects

- State#run! now ignores `input`
- State#run! now returns the block status (0) instead of the `output, next_state`
- State#start is no longer abstract (sorry)
- Setting `next_state` at the beginning of the `step` loop rather than the end

### DIFF
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -73,22 +73,11 @@ module Floe
       return Errno::EPERM if end?
 
       step_next
-
-      step_nonblock_start unless current_state.started?
-      return Errno::EAGAIN unless step_nonblock_ready?
-
-      step_nonblock_finish
+      current_state.run_nonblock!
     end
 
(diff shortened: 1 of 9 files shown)


kbrock:
updates:
- converted timeout syntax to your preferred `wait until run` style
- successfully moved start/finish into state
- moved more workflow state operations into state
- state#start now has a super implementation (you may not like this)

update:
- fixed hash rocket and guard clause cops

kbrock:
update:
- rebased
- changed to store strings

agrare (approved):

PR #99, opened 2023-09-21
RunnerContext should use string keys

agrare:
When serialized to a jsonb column the symbol keys are converted to string keys causing the values to appear missing when context is later deserialized

### DIFF
--- a/lib/floe/workflow/state.rb
+++ b/lib/floe/workflow/state.rb
@@ -41,7 +41,7 @@ module Floe
       end
 
       def finish
-        context.state["FinishedTime"] ||= Time.now.utc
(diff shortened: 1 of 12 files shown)


lib/floe/workflow/state.rb
```diff
@@ -41,7 +41,7 @@ def start(_input)
       end
 
       def finish
-        context.state["FinishedTime"] ||= Time.now.utc
+        context.state["FinishedTime"] ||= Time.now.utc.to_s
```
kbrock, line 44: do we want `.to_s` or `.iso8601`?

agrare: :+1: good call, fixed

PR #101, opened 2023-09-21
Cops

kbrock:
They are annoying and easy enough to fix

~~NOTE:~~
> \# ~~floe.gemspec~~
>  ~~spec.metadata['rubygems_mfa_required'] = 'true'~~

^ opted to not do that

### DIFF
--- a/floe.gemspec
+++ b/floe.gemspec
@@ -34,8 +34,4 @@ Gem::Specification.new do |spec|
   spec.add_dependency "kubeclient", "~>4.7"
   spec.add_dependency "more_core_extensions"
   spec.add_dependency "optimist", "~>3.0"
-
-  spec.add_development_dependency "manageiq-style"
-  spec.add_development_dependency "rspec"
-  spec.add_development_dependency "rubocop"
 end
--- a/Gemfile
+++ b/Gemfile
@@ -8,4 +8,7 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 # Specify your gem's dependencies in floe.gemspec
 gemspec
 
+gem "manageiq-style"
 gem "rake", "~> 13.0"
+gem "rspec"
+gem "rubocop"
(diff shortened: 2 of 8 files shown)


lib/floe/workflow/runner.rb
```diff
@@ -7,7 +7,7 @@ class Runner
 
       TYPES = %w[docker podman kubernetes].freeze
 
-      def initialize(_options = {})
+      def initialize(_options = {}) # rubocop:disable Style/RedundantInitialize
```
agrare, line 10: > Unnecessary disabling of Style/RedundantInitialize.

agrare: Rationale behind this, if the child classes are going to define initialize with arguments I like to define the interface at the base class.

floe.gemspec
```diff
@@ -35,7 +36,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "more_core_extensions"
   spec.add_dependency "optimist", "~>3.0"
 
-  spec.add_development_dependency "manageiq-style"
-  spec.add_development_dependency "rspec"
-  spec.add_development_dependency "rubocop"
+  spec.add_development_dependency "manageiq-style" # rubocop:disable Gemspec/DevelopmentDependencies
+  spec.add_development_dependency "rspec"          # rubocop:disable Gemspec/DevelopmentDependencies
+  spec.add_development_dependency "rubocop"        # rubocop:disable Gemspec/DevelopmentDependencies
```
agrare, line 41: @kbrock we could actually move these to the Gemfile, what do you think?

kbrock: I kinda like them where they are, but yea. I'll do that.

agrare: I don't really have an opinion either way but if it is a new cop that we're not sure we want wr could vote in manageiq-style

lib/floe/workflow/runner.rb
kbrock: I understood.

Locally it reports an error here (along with a bunch of "outdated" issues with style)
On the server, it seems ok. confused

floe.gemspec
kbrock: I moved to gemfile. easier to just go with the flow on this one

lib/floe/workflow/runner.rb
kbrock: I removed this rubocop disable

agrare (approved):

PR #102, opened 2023-09-22
only provide output once workflow is complete

kbrock:
### Before

`workflow#output` was available after every completed step

### After

`workflow#output` is only available after the workflow is completed

### DIFF
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -89,7 +89,7 @@ module Floe
     end
 
     def output
-      context.output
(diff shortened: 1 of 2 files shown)


Fryguy:
I'm not sure I understand this - don't we have output all along since each task produces new output?

kbrock:
We populate `output` in the workflow once the workflow is complete

agrare:
> don't we have output all along since each task produces new output?

We have each state's input and output while the workflow is running

PR #105, opened 2023-09-28
add Task#TimeoutSeconds{,Path}

kbrock:
Overview:

When there is an issue with the resource name, kubernetes runs forever.
I added timeout support to handle these issues

Highlights:

- `Runner#output,status` uses cached value after the container stops running.
- `Runner#mark_status` lets `Task` override the status 
- add `Task#TimeoutSeconds`
- add `Task#TimeoutSecondsPath`


kbrock:
update:
- fixed syntax error in timeoutpath section (refactoring FTW)

kbrock:
update:

- fixed hashrocket

kbrock:
@agrare Could you verify the first 2 commits.

1. Is it true that we only want to ask for the output after the task is complete?
2. If this is not true, then we need another way to change the runner context to state that there is an error / cause.
3. Even if that was true, do we want to set those values in the context, or do you feel setting the Error, Cause in the context are better
4. Does it make sense to have context more consistent with a tristate success (true, false, nil), Error, Cause, and Output column?

Due to the stubbing, it is a little tricky to test these values

kbrock:
WIP: want to discuss the first bullet "caching" before merging
Other than that, I'm happy with these changes

lib/floe/workflow/runner/docker.rb
```diff
@@ -73,8 +75,12 @@ def success?(runner_context)
...
 
         def output(runner_context)
-          output = docker!("logs", runner_context["container_ref"], :combined_output => true).output
-          runner_context["output"] = output
+          runner_context["output"] ||= docker!("logs", runner_context["container_ref"], :combined_output => true).output
+        end
+
+        def mark_status(runner_context, type, error: true)
+          runner_context["container_state"] = {"Running" => false, "ExitCode" => error ? 999 : 0}
+          runner_context["output"] = type
```
kbrock, line 83: this is defining the internal values into the `runner_context`.

Instead, do we want to define `Error`, `Cause`, and `Output?
Or do we prefer to define these in the `Context#state` and skip the `runner`/`runner_context` all together?

kbrock:
I rebased.

WIP: need to validate this on all 3 runtimes
WIP: need to determine where timeout lives (in runtime or task)
WIP: if it does belong in Task, need to determine the best way to override the runtime.

PR #106, opened 2023-09-28
[test] automatically add success states to payload

kbrock:
To have a complete workflow, we need to provide states for each of the next values.
But lots of the time, we are not concerned with the next states, but rather the state under test.

This automatically adds the next states so the workflow is complete, but minimally defined in tests.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -60,6 +60,17 @@ RSpec.configure do |config|
   def make_payload(states)
     start_at ||= states.keys.first
 
+    # add extra states to make sure the payload is complete
+    next_values = []
+    states.each_value do |value|
+      next_values << value["Next"]
+      if value["Type"] == "Choice"
+        next_values << value["Default"]
+        next_values += value["Choices"].map { |choice| choice["Next"] }
+      end
+    end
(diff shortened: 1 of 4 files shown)


kbrock:
This only really comes into play after we add validation.

There are 2 camps:
1. tests should define everything correctly
2. test should define as little as possible

I was hoping for option 2 but to have correct and complete workflows.

Fryguy:
I think there might be times where we don't want this, such as if we want to intentionally test broken workflows.  I wonder if it's cleaner to have a `include_success_states: true` parameter or something so it's opt-in or opt-out?

FWIW, I'm personally in camp 1, but I understand the appeal of camp 2.  I'm not necessarily opposed to it though.  I'll defer to @agrare on what is preferable.

agrare:
IMO think it is going to be very confusing to look at a test, see the payload that is used to create the workflow, and get "magic" extra states.  I think we should define the exact payload that we want and if we want to reduce repetition could we go with some factory_bot factories maybe?

kbrock:
Yea, https://github.com/ManageIQ/floe/pull/125 had to skip these helper methods for this reason.
But these are the only tests that I feel would need to be invalid.

This is not worth a debate.
We can revisit in the future but I'm sold on just going with camp 1

PR #107, opened 2023-09-28
[test] test Runner#run_async! and not Runner#run!

kbrock:
The main code runs `Runner#run_async!` but the tests go through `run!` which is a little different.

This changes the tests to testing `run_async!`, the methods that our code uses.

followups will obviously be

- Change `Task` tests to not completely stub `Runner`
- Drop `Runner#run!`
- Drop `Workflow` methods that are blocking (unsure)


kbrock:
update:
- rubocop fixes

agrare (approved):

PR #108, opened 2023-09-28
Add Fail#CausePath and Fail#ErrorPath

kbrock:
- Add `Fail` parameters `CausePath` and `ErrorPath`
- output contains `{Error:, Cause:}`

AWS gave a great example for `ErrorPath` and `CausePath`
This is used to convey a failed task
error/cause got into the final output

### DIFF
--- a/lib/floe/workflow/state.rb
+++ b/lib/floe/workflow/state.rb
@@ -92,6 +92,25 @@ module Floe
       def finished?
         context.state.key?("FinishedTime")
       end
+
+      private
+
+      # Use a payload value or hardcoded path.
+      #
+      # @param [Context] context     context
+      # @param [Hash|String] input   state input
+      # @param [Object] value        hardcoded value from the payload
+      # @param [Path|Nil] value_path path to the value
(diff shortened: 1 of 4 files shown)


lib/floe/workflow/state.rb
```diff
@@ -92,6 +92,25 @@ def ready?
...
+      # @yield [String]              block to convert path fetched string into proper datatype
+      # @returns [Object]            value derived from hardcoded value or path
+      def value_or_path(context, input, value = nil, path:)
+        if path
+          value = path.value(context, input)
+          block_given? ? yield(value) : value
+        else
+          value
+        end
+      end
```
agrare, line 113: Hm I don't love methods that behave completely differently based on essentially a boolean input param, seems like this could be completely replaced by `@error_path.nil? ? error : @error_path.value(context, input)` in the caller and it is more obvious what is happening.

agrare (approved):
This is a good fix for our Fail state missing some important attributes so I'll merge as is

PR #109, opened 2023-09-28
Drop Runner#run!

kbrock:
### Overview

The various layers (`Workflow`, `State`, `Task`, and `Runner`) have quite a few methods forthat do blocking and non-blocking actions.

Feel like having only non-blocking methods would simplify the code.

If we want to provide blocking options to the outside world, that can be done in workflow using the current pattern in place.

depends:
- [x] #113 

Remove `Runner#run!`

`Task#run_nonblock!` only calls `Runner#run_nonblock!`. Since nothing calls into `Runner#run!` there is no need to keep this around.

questions:

- Remove `State#run` -- api will be `run_nonblock`

I'm thinking we'll keep `Workflow#step`, but it will call `run_nonblock` / wait combo.

### DIFF
--- a/lib/floe/workflow/runner/docker.rb
+++ b/lib/floe/workflow/runner/docker.rb
@@ -13,23 +13,6 @@ module Floe
           @network = options.fetch("network", "bridge")
         end
 
-        def run!(resource, env = {}, secrets = {})
-          raise ArgumentError, "Invalid resource" unless resource&.start_with?("docker://")
-
-          image = resource.sub("docker://", "")
-
-          secrets_file = nil
-          if secrets && !secrets.empty?
-            secrets_file = create_secret(secrets)
-          end
-
-          output = run_container(image, env, secrets_file)
-
-          {"exit_code" => 0, "output" => output}
-        ensure
-          cleanup({"secrets_ref" => secrets_file})
-        end
-
         def run_async!(resource, env = {}, secrets = {})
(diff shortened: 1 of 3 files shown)


Fryguy:
it's used by the CLI, I believe: https://github.com/ManageIQ/floe/blob/master/exe/floe#L37

kbrock:
`floe` calls `Workflow#run!` and `WorkflowInstance` call `Workflow#run_nonblock!`.

`Workflow` calls into `State#run_nonblock!` or if it does call `State#run`, that calls `State#run_nonblock!`. From there, `Task#run_nonblock!` calls `Runner#run_nonblock!`.

Nothing calls `Runner#run_nonblock!`

But since the code is synchronous, we may want to keep it around. If we decide to do that, I will probably refactor them to reduce code duplication. Just deleting the method seems like the cheapest route.

kbrock:
WIP: This is only wip to delay the merge until after this release. I plan on un-WIPing it and leaving code as is.

kbrock:
update:
- rebase

kbrock:
update:
- rebase, it now only contains the 1 commit of interest

agrare (approved):
So originally I left the "blocking" mode in for the CLI because it was more efficient than the async+sleep/poll but as we move towards blocking event-driven updates with `Workflow#wait` it should be negligible and there is a decent amount of simplification we can get here so I think it is worth it.

PR #110, opened 2023-09-29
Task#Catcherr puts an "Error" into the output

kbrock:
https://states-language.net/#error-output

---
I'm pretty sure this should use the raw input and not the effective/processed input

picture form:

https://docs.aws.amazon.com/step-functions/latest/dg/concepts-input-output-filtering.html

text form:

https://states-language.net/#using-inputpath-parameters-resultselector-resultpath-and-outputpath

item 4 says that the "raw input" is used instead of the "effective input"

### DIFF
--- a/lib/floe/workflow/states/task.rb
+++ b/lib/floe/workflow/states/task.rb
@@ -104,6 +104,7 @@ module Floe
           raise error if catcher.nil?
 
           context.next_state = catcher.next
+          context.output     = catcher.result_path.set(context.input, {"Error" => error})
         end
 
         def process_input(input)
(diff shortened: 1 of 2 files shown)


lib/floe/workflow/states/task.rb
```diff
@@ -113,7 +114,7 @@ def process_input(input)
         end
 
         def process_output!(results)
-          output = process_input(context.state["Input"])
+          output = context.input.dup
```
agrare, line 117: We still need to use `process_input` or something like it to apply the `InputPath` and the `Parameters`

kbrock: I updated the description above (to include some of the commit comments)

From what I can tell, the output is based upon the raw input and not the processed input.

agrare (approved):

Issue #111, opened 2023-10-03
Fix Task output and error handling for docker resources


agrare:
When a docker resource is completed we use the container logs to feed the output / error handling of the Task state.

Currently the entirety of stdout+stderr is considered but this breaks down quickly when you consider that other libraries might be logging to stdout/stderr and also that you want to log progress during a long running container and only print output at the end.

We are suggesting that a delimiter be printed once a container has completed and is printing output for task output consideration and everything after that point will be considered for task output.

agrare:
cc @kbrock

kbrock:
We also have the need for the docker container to log output.

As mentioned elsewhere, a typical lambda process has 3 output channels: `stdout`, `stderr`, and `data` (via `return`).

We have condensed down to a single channel. While `stderr` and `stdout` can be consolidated to a `logging` channel, we still need to separate out the `data`.

We also need to know if there was a success of a failure.
If the process is connecting to an http endpoint, printing out the stack trace 

```
connecting to provider
fetching values
retrieved values
__FLOE__OUTPUT__
{"value": "Yay"}
```

Not sure the best way to include an error and a cause, but we can possibly choose a delimiter, or use the same delimiter and use a common format.

```
error on line 18
called from 15
called from 12
__FLOE__OUTPUT__
{"Error": "Timeout", "Cause": "Could not connect to api within 1 minute"}
```

or

```
...
__FLOE__ERROR__OUTPUT__
{"Error": "Timeout", "Cause": "Could not connect to api within 1 minute"}
```

Fryguy:
> We have condensed down to a single channel.

Not sure I understand this - docker supports stderr and stdout in the logging. Does `docker logs` not support that split?  Or is this a podman or kubernetes thing?  Just want to make sure I understand this comment correctly before moving on.

agrare:
docker and podman have stdout/stderr split up, kubernetes doesn't

PR #112, opened 2023-10-03
Fix task output handling by adding an output marker

agrare:
Only consider output after the last occurrence of a pre-defined output marker.

Depends on:
- [x] https://github.com/ManageIQ/floe/pull/116
- [x] https://github.com/ManageIQ/floe/pull/115

Fixes https://github.com/ManageIQ/floe/issues/111

### DIFF
--- a/lib/floe/workflow/runner.rb
+++ b/lib/floe/workflow/runner.rb
@@ -5,7 +5,8 @@ module Floe
     class Runner
       include Logging
 
-      TYPES = %w[docker podman kubernetes].freeze
+      TYPES         = %w[docker podman kubernetes].freeze
+      OUTPUT_MARKER = "__FLOE_OUTPUT__\n"
 
       def initialize(_options = {})
       end
--- a/lib/floe/workflow/states/task.rb
+++ b/lib/floe/workflow/states/task.rb
@@ -29,10 +29,10 @@ module Floe
 
         def start(input)
           super
-          input = input_path.value(context, input)
-          input = parameters.value(context, input) if parameters
 
+          input          = process_input(input)
           runner_context = runner.run_async!(resource, input, credentials&.value({}, workflow.credentials))
+
           context.state["RunnerContext"] = runner_context
         end
 
@@ -41,13 +41,15 @@ module Floe
         end
 
         def finish
-          results = runner.output(context.state["RunnerContext"])
+          output = runner.output(context.state["RunnerContext"])
 
           if success?
-            context.state["Output"] = process_output!(results)
+            output = parse_output(output)
+            context.state["Output"] = process_output!(output)
             context.next_state      = next_state
           else
-            retry_state!(results) || catch_error!(results) || fail_workflow!(results)
+            error = parse_error(output)
+            retry_state!(error) || catch_error!(error) || fail_workflow!(error)
           end
 
           super
@@ -81,7 +83,7 @@ module Floe
         end
 
         def retry_state!(error)
-          retrier = find_retrier(error)
+          retrier = find_retrier(error["Error"]) if error
           return if retrier.nil?
 
           # If a different retrier is hit reset the context
@@ -100,17 +102,17 @@ module Floe
         end
 
         def catch_error!(error)
-          catcher = find_catcher(error)
+          catcher = find_catcher(error["Error"]) if error
           return if catcher.nil?
 
           context.next_state = catcher.next
-          context.output     = catcher.result_path.set(context.input, {"Error" => error})
+          context.output     = catcher.result_path.set(context.input, error)
           true
         end
 
         def fail_workflow!(error)
           context.next_state     = nil
-          context.output         = {"Error" => error}.compact
+          context.output         = {"Error" => error["Error"], "Cause" => error["Cause"]}.compact
           context.state["Error"] = context.output["Error"]
         end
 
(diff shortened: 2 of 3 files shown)


lib/floe/workflow/runner/docker.rb
```diff
@@ -74,7 +74,13 @@ def success?(runner_context)
 
         def output(runner_context)
           output = docker!("logs", runner_context["container_ref"], :combined_output => true).output
-          runner_context["output"] = output
+
+          marker_location = output.rindex(OUTPUT_MARKER)
+          return if marker_location.nil?
```
kbrock, line 79: no marker = bad output. or assume all is output?

I guess if we don't require it, and then a gem gets updated and randomly outputs, it would be sporadic. (legit case you are seeing)

If we do require it, then we'll force this to be present, and that will be more resilient.

lib/floe/workflow/runner/docker.rb
```diff
@@ -74,7 +74,13 @@ def success?(runner_context)
...
+
+          output_location = marker_location + OUTPUT_MARKER.length
+
+          runner_context["output"] = output[output_location..]
```
kbrock, line 83: hmm.

```
runner_context["output"] = output.split(OUTPUT_MARKER).last
```

kbrock:
This looks good.
Do we want to solve errors now, or at a later time.

Example I see is we try and connect to the rest api, but it doesn't go through.

lib/floe/workflow/runner/docker.rb
agrare: Yeah I'd prefer we require it, that way we know for sure anything without it is "logs" at runtime

agrare:
>  Do we want to solve errors now, or at a later time?

I'd like to solve errors in this PR, added error handling with the same output and payload of `{"Error" => "ErrorClass", "Message" => "Longer Error Message"}`

The ErrorClass is used to match the catcher/retrier

spec/workflow/states/task_spec.rb
```diff
@@ -273,7 +280,11 @@
             .to receive(:run_async!)
             .with(payload["Resource"], input, nil)
 
-          expect { state.run!(input) }.to raise_error(RuntimeError, "Exception")
+          state.run!(input)
+
+          expect(ctx.next_state).to     be_nil
+          expect(ctx.status).to         eq("failure")
+          expect(ctx.state["Error"]).to eq("Exception")
```
kbrock, line 287: can we test the `output` here instead of `state["Error"]`?

I think I'm ready to retire those special fields

lib/floe/workflow/states/task.rb
```diff
@@ -100,11 +102,22 @@ def retry_state!(error)
         end
 
         def catch_error!(error)
-          catcher = find_catcher(error)
-          raise error if catcher.nil?
+          catcher = find_catcher(error["Error"]) if error
+          return if catcher.nil?
 
           context.next_state = catcher.next
           context.output     = catcher.result_path.set(context.input, {"Error" => error})
```
kbrock, line 109: Think this should be:

```suggest
context.output     = catcher.result_path.set(context.input, error)
```

kbrock (commented):
very excited about this

lib/floe/workflow/runner.rb
```diff
@@ -5,7 +5,8 @@ class Workflow
     class Runner
       include Logging
 
-      TYPES = %w[docker podman kubernetes].freeze
+      TYPES         = %w[docker podman kubernetes].freeze
+      OUTPUT_MARKER = "__FLOE_OUTPUT__\n"
```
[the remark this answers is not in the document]
agrare: rubocop warned about freezing static objects so I removed it.

> ❗ - [Line 9](https://github.com/agrare/floe/blob/389d699d70c58f5f4518be96409f325c38bed0cc/lib/floe/workflow/runner.rb#L9), Col 23 - [Style/RedundantFreeze](https://docs.rubocop.org/rubocop/1.28/cops_style.html#styleredundantfreeze) - Do not freeze immutable objects, as freezing them has no effect.

lib/floe/workflow/runner/docker.rb
```diff
@@ -74,7 +74,9 @@ def success?(runner_context)
 
         def output(runner_context)
           output = docker!("logs", runner_context["container_ref"], :combined_output => true).output
-          runner_context["output"] = output
+          return unless output.index(OUTPUT_MARKER)
```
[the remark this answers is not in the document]
agrare: > Do we want to raise an error here so the user knows something is wrong?

No because I don't want to assume that the container has finished at this point, it could be running and simply not printed all output yet.

spec/workflow/states/task_spec.rb
agrare: :+1: good call will do

Fryguy:
Would this be on the docker container itself to be sure it outputs that marker?

lib/floe/workflow/runner.rb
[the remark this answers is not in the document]
kbrock: aah

```
# frozen_string_literal: true
```

lib/floe/workflow/runner/docker.rb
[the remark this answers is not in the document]
kbrock: I didn't think we called `output` until after the container has been marked as done
and `output` was only called once. never an intermediate value.

Hmm... that doesn't make debugging that useful

Fryguy:
Yeah, I was worried about this - I was trying for a while to avoid forcing the output to be a certain way 🤔   I was originally hoping we could do it the lambda way, but then it forces us to support specific languages, which goes against the entire idea of the end user using whatever they want.

> We also didn't want to say e.g. the "last" line is the output since that is very limiting.

My original thought was what if we just used any trailing JSON?  Or perhaps any JSON sent to stdout we pull in and ignore the rest?  Not sure.

Fryguy:
Similarly on that thought, I think we already require a JSON output, so it's possible we can make the marker part of the JSON output itself (thinking out loud here)  e.g.

```json
{
  "_FLOE_OUTPUT_": true,
  ...
}
```

agrare:
> I think we already require a JSON output

Technically we wrap non-json output in `{"results => output}` if it isn't a json object but I always disliked that.  I'd be fine dropping that and requiring JSON output.

agrare:
> My original thought was what if we just used any trailing JSON?

Yeah this was my first take one it, it is definitely the simplest.  Some issues that led to the delimeter were simplicity in dealing with multi-line output (even a pretty-json output) and it makes it clear even while a container is still running what is "output" vs "logging".

I would like to get this in since assuming all text written to stdout is output is not workable, could we agree on the last line if it is JSON is the task output?

PR #113, opened 2023-10-04
[test] Drop calls to blocking run/step

kbrock:
Our code calls into `run_nonblock!` and `step_nonblock!`

Changed tests to reflect our use case.

Also dropping the `sleep` stub. This is not being called, and since we don't want `sleep` calls to creep in, this will make it very obvious and catch the issue quickly.

---

Obvious next move is to drop all `sleep` statements and blocking calls.
We will actually keep the sleep in `State#run_wait`. `Workflow#run` and possibly `Workflow#wait` will use this method.

### DIFF
--- a/spec/workflow/states/choice_spec.rb
+++ b/spec/workflow/states/choice_spec.rb
@@ -32,10 +32,10 @@ RSpec.describe Floe::Workflow::States::Choice do
     expect(state.end?).to eq(false)
   end
 
-  describe "#run!" do
+  describe "#run_nonblock!" do
     context "with a missing variable" do
       it "raises an exception" do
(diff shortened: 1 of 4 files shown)


kbrock:
WIP: this is low priority
After the next release, I plan on un-WIP and leaving code as is.

kbrock:
update:
- rebased
- extracted `Task` changes into its own PR

agrare (approved):

kbrock:
un-wip: We released, seems like a good time to merge

kbrock:
update:
- rebase

agrare (approved):

PR #117, opened 2023-10-06
Workflow#wait_until

kbrock:
This exposes the `wait_until` for both Task (retry) and Wait. This can be used by the [workflow runner] for a better `deliver_on` value.

At a first glance, I feel it would be nice to hide this value.
After thinking about this, a `run_nonblock` stops due to 3 reasons:

1. The workflow has ended (success or failure)
2. The workflow is blocked on a Task container run
3. The state has requested a sleep (Task retry or standard Wait)

I don't think asking "how long until I can call you back" seems so bad.
Waiting on a container will not change/have a response (though we may want to put the hardcoded 1.second in the runner instead of the workflow provider). But the Wait and Task retrier will respond with a time when to call back.

[Workflow runner]: https://github.com/manageiq/manageiq-providers-workflows/blob/master/app/models/manageiq/providers/workflows/automation_manager/workflow_instance.rb#L93

### DIFF
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -96,6 +96,10 @@ module Floe
       context.ended?
     end
 
+    def wait_until
+      context.wait_until
+    end
+
     def current_state
       @states_by_name[context.state_name]
     end
--- a/lib/floe/workflow/context.rb
+++ b/lib/floe/workflow/context.rb
@@ -73,6 +73,10 @@ module Floe
         end
       end
 
+      def wait_until
+        state["WaitUntil"] && Time.parse(state["WaitUntil"])
+      end
+
       def state=(val)
         @context["State"] = val
       end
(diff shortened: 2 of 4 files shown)


kbrock:
update:
- `1.second` => `1` (dropped active support)

kbrock:
@agrare Not sure if you saw my updated reasoning above. I feel hiding this wait value is too academic.

agrare:
> I don't think asking "how long until I can call you back" seems so bad.

I think it is impossible to guess for Task which makes it only helpful for Wait.  My primary hesitation on merging this is that if we then want to remove it that would constitute a breaking API change and necessitate a major version bump.

IMO the more efficient way of handling this is by using update-driven watches to drive the `Floe::Workflow.wait` method which will allow blocking/waking down to the millisecond.

kbrock:
This should not be too hard to write again

PR #118, opened 2023-10-06
drop value_or_path

kbrock:
No need for `value_or_path`
The alternative is just as easy to read
and even a hint shorter.

When I was writing this code, it was easier to reason through with this abstraction.
Now, it is no longer necessary

---

@agrare you seemed to want to get away from this.
Unfortunate that this adds some "complexity" to `Wait#start`

### DIFF
--- a/lib/floe/workflow/states/wait.rb
+++ b/lib/floe/workflow/states/wait.rb
@@ -29,8 +29,8 @@ module Floe
           context.output     = output_path.value(context, input)
           context.next_state = end? ? nil : @next
           wait(
-            :seconds => value_or_path(context, input, @seconds, :path => @seconds_path, &:to_i),
-            :time    => value_or_path(context, input, @timestamp, :path => @timestamp_path)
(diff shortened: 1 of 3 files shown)


lib/floe/workflow/states/wait.rb
```diff
@@ -29,8 +29,8 @@ def start(input)
           context.output     = output_path.value(context, input)
           context.next_state = end? ? nil : @next
           wait(
-            :seconds => value_or_path(context, input, @seconds, :path => @seconds_path, &:to_i),
-            :time    => value_or_path(context, input, @timestamp, :path => @timestamp_path)
+            :seconds => @seconds_path ? @seconds_path.value(context, input).to_i : @seconds,
+            :time    => @timestamp_path ? @timestamp_path.value(context, input) : @timestamp
           )
```
kbrock, line 34: If I remember correctly. Before introducing this method I pulled this out into a separate method to avoid this complexity complaint

Let me know if you want to pull that out.

agrare (approved):

PR #120, opened 2023-10-06
bin/floe converts input to json

kbrock:
The Context expects input to be a json object.
bin/floe is passing in a raw string to the Context.

BEFORE:

If there is an error in the Task, the result_path is not able to apply to the input.
This causes an error:

```
lib/floe/workflow/reference_path.rb:35:in `set': undefined method `merge!' for "{\\"foo\\": 2}":String (NoMethodError)

          result.merge!(value)
                ^^^^^^^
  from lib/floe/workflow/states/task.rb:114:in `catch_error!'
  from lib/floe/workflow/states/task.rb:55:in `finish'
  from lib/floe/workflow/state.rb:51:in `run_nonblock!'
  from lib/floe/workflow.rb:76:in `step_nonblock'
  from lib/floe/workflow.rb:63:in `step'
  from lib/floe/workflow.rb:58:in `run!'
  from exe/floe:37:in `<top (required)>'
```

After:

An error is able to be merged with the input

```
{"Error"=>"FailStateError", "Cause"=>"No Matches!"}
```

### DIFF
--- a/exe/floe
+++ b/exe/floe
@@ -31,7 +31,7 @@ runner_options = opts[:docker_runner_options].to_h { |opt| opt.split("=", 2) }
 
 Floe::Workflow::Runner.docker_runner = runner_klass.new(runner_options)
 
-context = Floe::Workflow::Context.new(:input => opts[:input])
+context = Floe::Workflow::Context.new(:input => JSON.parse(opts[:input]))
 workflow = Floe::Workflow.load(opts[:workflow], context, opts[:credentials])
 
 workflow.run!


exe/floe
```diff
@@ -31,7 +31,7 @@ runner_options = opts[:docker_runner_options].to_h { |opt| opt.split("=", 2) }
 
 Floe::Workflow::Runner.docker_runner = runner_klass.new(runner_options)
 
-context = Floe::Workflow::Context.new(:input => opts[:input])
+context = Floe::Workflow::Context.new(:input => JSON.parse(opts[:input]))
```
agrare, line 34: @kbrock what do you think about moving this into the `Context` initializer?  We already do a similar `context = JSON.parse(context) if context.kind_of?(String)` for the context if it is passed in.  Otherwise all callers would have to do this e.g. https://github.com/ManageIQ/manageiq-providers-workflows/blob/master/app/models/manageiq/providers/workflows/automation_manager/workflow.rb#L14

kbrock: Is a string valid json input?
Can we assume we only want a hash (though I guess array could work)

agrare: > When a state machine is started, the caller can provide an initial [JSON text](https://tools.ietf.org/html/rfc8259#section-2) as input, which is passed to the machine's start state as input.

I believe we can assume that input must be JSON per the spec

agrare:
Merged https://github.com/ManageIQ/floe/pull/122

PR #122, opened 2023-10-06
Json parse Context#input if it is a string

kbrock:
alt to #120


agrare:
> When a state machine is started, the caller can provide an initial [JSON text](https://tools.ietf.org/html/rfc8259#section-2) as input, which is passed to the machine's start state as input.

I believe we can assume that input must be JSON per the spec

PR #127, opened 2023-10-06
exe/floe output

kbrock:
- add option `exe/floe --no-log` to not print logging to standard out (defaults to logging)
- add return status based upon the workflow running successfully

### DIFF
--- a/exe/floe
+++ b/exe/floe
@@ -11,12 +11,13 @@ opts = Optimist.options do
   opt :credentials, "JSON payload with credentials",           :default => "{}"
   opt :docker_runner, "Type of runner for docker images",      :default => "docker"
   opt :docker_runner_options, "Options to pass to the runner", :type => :strings
+  opt :log, "Show the log",                                    :type => :boolean, :default => true
 end
 
 Optimist.die(:docker_runner, "must be one of #{Floe::Workflow::Runner::TYPES.join(", ")}") unless Floe::Workflow::Runner::TYPES.include?(opts[:docker_runner])
 
 require "logger"
-Floe.logger = Logger.new($stdout)
+Floe.logger = Logger.new($stdout) if opts[:log]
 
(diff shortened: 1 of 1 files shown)


Fryguy:
@kbrock Just curious what problem you're trying to solve with the --no-log ?  The exit 0/1 makes total sense.

exe/floe
```diff
@@ -11,12 +11,13 @@ opts = Optimist.options do
...
   opt :docker_runner, "Type of runner for docker images",      :default => "docker"
   opt :docker_runner_options, "Options to pass to the runner", :type => :strings
+  opt :log, "Show the log",                                    :type => :boolean, :default => true
 end
 
 Optimist.die(:docker_runner, "must be one of #{Floe::Workflow::Runner::TYPES.join(", ")}") unless Floe::Workflow::Runner::TYPES.include?(opts[:docker_runner])
 
 require "logger"
-Floe.logger = Logger.new($stdout)
+Floe.logger = Logger.new($stdout) if opts[:log]
```
agrare, line 20: Is this any different than `floe ... >/dev/null` ?

kbrock: All of the state changes and logging goes away.
But the final output `context.output` / `workflow.output` is displayed.

This is not totally necessary

kbrock:
I have a couple of integration tests locally. Mostly with failures in various places.
So locally I commented out this line to not see so much gook on the screen

It is not needed

Fryguy:
Definitely make a PR for the exit 1/0 - that's a no-brainer

exe/floe
kbrock: But with typical programs, I'd expect `--verbose` to display these values.

kbrock:
punting.
If it becomes an issue in the future, I'll open another one with `--verbose` and changing the log level (I liked that suggestion)

PR #131, opened 2023-10-12
Add ability to pass workflow service account to kube runner

Fryguy:
@agrare @kbrock Please review

cc @bdunne

### DIFF
--- a/lib/floe/workflow/runner/kubernetes.rb
+++ b/lib/floe/workflow/runner/kubernetes.rb
@@ -35,6 +35,8 @@ module Floe
 
           @namespace = options.fetch("namespace", "default")
 
+          @task_service_account = options["task_service_account"]
+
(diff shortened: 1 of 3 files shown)


lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -35,6 +35,8 @@ def initialize(options = {})
 
           @namespace = options.fetch("namespace", "default")
 
+          @workflow_service_account = options["workflow_service_account"]
```
Fryguy, line 38: Naming is hard - couldn't figure out a good name here, but if you have other ideas, please mention.

Fryguy: Agreed on task_service_account, since this is specifically for the tasks that are launched.

PR #132, opened 2023-10-13
Prefix pod names with 'floe-'

Fryguy:
In Kubernetes, when running in namespace, the "random" nature of the pod
name being based on the image name causes these to appear in a pod
listing in random locations. This change to prefix with 'floe-' allows
all of these pods to appear next to each other aiding in identification.
This can help with things like finding pods that are failing to be deleted.

This commit also changes the docker runner and podman runner to use
proper names that match those we are using for kubernetes, for similar
reasons such as in `docker ps`.

@agrare @kbrock Please review.

I have not tested this with podman locally, but I did test with kubernetes and docker.


lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -113,7 +104,7 @@ def pod_spec(name, image, env, secret = nil)
             :spec       => {
               :containers    => [
                 {
-                  :name  => container_name(image),
+                  :name  => image_name(image),
```
Fryguy, line 107: This one was interesting to me - not sure why we couldn't also use the pod name here?  Just curious why this was specially called out.  Regardless, I updated the spec to call out this difference.

lib/floe/workflow/runner/docker_mixin.rb
```diff
@@ -0,0 +1,18 @@
...
+      module DockerMixin
+        def image_name(image)
+          image.match(%r{^(?<repository>.+/)?(?<image>.+):(?<tag>.+)$})&.named_captures&.dig("image")
+        end
+
+        def container_name(image)
+          name = image_name(image)
+          raise ArgumentError, "Invalid docker image [#{image}]" if name.nil?
+
+          "floe-#{name}-#{SecureRandom.uuid}"
```
Fryguy, line 13: I think there's technically a prior bug here since the pod_name can only be 63 characters, and it's possible to have an image name longer than (63 - 2 - 4 - 36 =) 21 characters.  Before this change, we could afford 26 characters, which is still pretty small.

I don't think we need a full UUID either - thinking we might be able to use, say, 8 or 12 random hex bytes.

lib/floe/workflow/runner/kubernetes.rb
agrare: Typically what I've seen on kubernetes the pod has the "unique" suffix and the container name in the pod is the simple image name.  The difference with kubernetes and docker/podman is that in kubernetes we're creating a pod which contains a container and on docker/podman we're creating the container directly there is no other layer above it.

agrare: Probably `floe-hello-world` would be the most consistent (e.g. `1-ems-metrics-processor-968dbbd67-sg886` container name is `1-ems-metrics-processor`) unfortunately this is neither container_name nor image_name

agrare (approved):

lib/floe/workflow/runner/kubernetes.rb
Fryguy: Sounds good - as discussed, let's merge and I'll get it in the follow up for the vairous weird docker names

PR #134, opened 2023-10-16
Fix issue where certain docker image names cannot be pod names

### LINKED ISSUE #133: Pod names might overload the 63 character limit
Extracted from https://github.com/ManageIQ/floe/pull/132#discussion_r1358348735

>  I think there's technically a prior bug here since the pod_name can only be 63 characters, and it's possible to have an image name longer than (63 - 2 - 4 - 36 =) 21 characters. Before this change, we could afford 26 characters, which is still pretty small.
>
> I don't think we need a full UUID either - thinking we might be able to use, say, 8 or 12 random hex bytes.

`SecureRandom.uuid` is overly large for what we require, so I'm thinking we can use something significantly smaller, but wanted to discuss before I wrote a PR.

I'm thinking something like `SecureRandom.urlsafe_base64(8)` which should give us more than enough entropy in a tiny payload.

method | bits | chars | notes|
--- | --- | --- | ---
`SecureRandom.uuid` | 128 | 36 | 36 = 32 chars + 4 `-`
`SecureRandom.urlsafe_base64(22)` | 128 | 22
`SecureRandom.urlsafe_base64(11)` | 64 | 11 | I think this is what YouTube uses
`SecureRandom.hex(6)` | 48 | 12
`SecureRandom.urlsafe_base64(8)` | 48 | 8
`SecureRandom.hex(4)` | 32 | 8

We don't actually need uuid's full 128 bits of entropy (2<sup>128</sup> or 340 undecillion), since we're just trying to avoid collisions on veru short-lived things. IMO 48 bits of entropy (2<sup>48</sup> or 281 trillion) is _way_ more than enough.

@agrare @kbrock Just want to come to an agreement here, then I'll write up the PR.

Fryguy:
- Handles image names longer than 63 characters
- Handles image names with `.` and `_` characters

Fixes #133

@agrare Please review.

### DIFF
--- a/spec/workflow/runner/kubernetes_spec.rb
+++ b/spec/workflow/runner/kubernetes_spec.rb
@@ -35,7 +35,7 @@ RSpec.describe Floe::Workflow::Runner::Kubernetes do
         :spec       => hash_including(
           :containers => [
             hash_including(
-              :name  => "hello-world",
+              :name  => "floe-hello-world",
               :image => "hello-world:latest"
             )
           ]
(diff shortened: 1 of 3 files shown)


spec/workflow/runner/kubernetes_spec.rb
```diff
@@ -46,6 +47,52 @@
...
+        :kind       => "Pod",
+        :apiVersion => "v1",
+        :metadata   => {
+          :name      => a_string_matching(/^floe-this-is-a-very-long-image-name-way-longer-than-sh\-\h{8}$/).and(a_string_matching(kube_pod_name_rules)),
+          :namespace => "default"
+        },
+        :spec       => hash_including(
+          :containers => [
+            hash_including(
+              :name  => "this-is-a-very-long-image-name-way-longer-than-should-be-allowed",
```
Fryguy, line 61: I forgot to check if the container name needs to be changed as well.

agrare (approved):

PR #135, opened 2023-10-19
Handle ImagePullErr/ImagePullBackOff as errors

agrare:
When a kubernetes pod fails to pull an image the pod stays in "Pending" indefinitely.  This causes floe to "hang" waiting for the pod to start but it never will.

### DIFF
--- a/lib/floe/workflow/runner/kubernetes.rb
+++ b/lib/floe/workflow/runner/kubernetes.rb
@@ -6,8 +6,10 @@ module Floe
       class Kubernetes < Floe::Workflow::Runner
         include DockerMixin
 
-        TOKEN_FILE   = "/run/secrets/kubernetes.io/serviceaccount/token"
-        CA_CERT_FILE = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+        TOKEN_FILE      = "/run/secrets/kubernetes.io/serviceaccount/token"
+        CA_CERT_FILE    = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+        RUNNING_PHASES  = %w[Pending Running].freeze
+        FAILURE_REASONS = %w[ImagePullBackOff ErrImagePull].freeze
 
         def initialize(options = {})
           require "awesome_spawn"
@@ -66,7 +68,17 @@ module Floe
         end
 
         def running?(runner_context)
-          %w[Pending Running].include?(runner_context.dig("container_state", "phase"))
+          phase              = runner_context.dig("container_state", "phase")
+          container_statuses = runner_context.dig("container_state", "containerStatuses") || []
+          container_reasons  = container_statuses.map { |status| status.dig("state", "waiting", "reason") }.compact
+
+          return false if RUNNING_PHASES.none?(phase)
+          # If an image failed to pull then the phase will be Pending but the
+          # container status will be waiting with a reason of ImagePullBackOff
+          # or ErrImagePull
+          return false if (container_reasons & FAILURE_REASONS).any?
+
+          true
         end
 
         def success?(runner_context)
--- a/lib/floe/workflow/runner/kubernetes.rb
+++ b/lib/floe/workflow/runner/kubernetes.rb
@@ -85,7 +85,7 @@ module Floe
           runner_context["output"] =
             if container_failed?(runner_context)
               failed_state = failed_container_states(runner_context).first
-              "#{failed_state["reason"]}: #{failed_state["message"]}"
+              {"Error" => failed_state["reason"], "Cause" => failed_state["message"]}
             else
               kubeclient.get_pod_log(runner_context["container_ref"], namespace).body
             end
--- a/lib/floe/workflow/runner/kubernetes.rb
+++ b/lib/floe/workflow/runner/kubernetes.rb
@@ -9,7 +9,7 @@ module Floe
         TOKEN_FILE      = "/run/secrets/kubernetes.io/serviceaccount/token"
         CA_CERT_FILE    = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
         RUNNING_PHASES  = %w[Pending Running].freeze
-        FAILURE_REASONS = %w[ImagePullBackOff ErrImagePull].freeze
+        FAILURE_REASONS = %w[CrashLoopBackOff ImagePullBackOff ErrImagePull].freeze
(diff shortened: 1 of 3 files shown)


lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -66,7 +68,17 @@ def status!(runner_context)
...
-          %w[Pending Running].include?(runner_context.dig("container_state", "phase"))
+          phase              = runner_context.dig("container_state", "phase")
+          container_statuses = runner_context.dig("container_state", "containerStatuses") || []
+          container_reasons  = container_statuses.map { |status| status.dig("state", "waiting", "reason") }.compact
+
+          return false unless RUNNING_PHASES.include?(phase)
+          # If an image failed to pull then the phase will be Pending but the
+          # container status will be waiting with a reason of ImagePullBackOff
+          # or ErrImagePull
+          return false unless (container_reasons & FAILURE_REASONS).empty?
```
kbrock, line 79: So it is not running, that means we will fetch a new status.
What `Error` will get assigned to the Task?

Fryguy:
@agrare Just for my clarity, what causes each of those error states?

lib/floe/workflow/runner/kubernetes.rb
agrare: This will return `false` if container_reasons includes any of the error failure states, and `success?` will return `false` since the phase isn't `Succeeded`.

Good point about the Task Error I should return the failure reason from `#output` if it is in this state.

agrare:
@Fryguy I reproduced by just typo'ing an image name, first you get `ErrImagePull` then after a period of time the pod will be in `ImagePullBackOff` so really the same root cause can be two different errors depending on when we check.

lib/floe/workflow/runner/kubernetes.rb
agrare: I think the `return false unless [].empty?` is needlessly difficult to grock and `return false if failures.any?` would read easier

Fryguy: more_core_extensions has this as defined as `include_any?` (or `include_none?` if you want the opposite) for readbaility.

lib/floe/workflow/runner/kubernetes.rb
```diff
@@ -66,7 +68,17 @@ def status!(runner_context)
...
-          %w[Pending Running].include?(runner_context.dig("container_state", "phase"))
+          phase              = runner_context.dig("container_state", "phase")
+          container_statuses = runner_context.dig("container_state", "containerStatuses") || []
+          container_reasons  = container_statuses.map { |status| status.dig("state", "waiting", "reason") }.compact
+
+          return false if RUNNING_PHASES.none?(phase)
+          # If an image failed to pull then the phase will be Pending but the
+          # container status will be waiting with a reason of ImagePullBackOff
+          # or ErrImagePull
+          return false if (container_reasons & FAILURE_REASONS).any?
```
Fryguy, line 79: So, with more_core_extensions this would be 

```suggestion
          return false if container_reasons.include_any?(FAILURE_REASONS)
```

kbrock:
@agrare could you add a test that shows the `output` will end up as `State.Timeout` or what ever

Fryguy:
Not sure I understand how this works...does the fact that it's no longer `running?` trigger some next step to say it's an error?

agrare:
> Not sure I understand how this works...does the fact that it's no longer running? trigger some next step to say it's an error?

The problem with these "failed" containers is that the pod is Pending, we were considering Pending to be a running state since that is the same state a pod which just hasn't started yet is in.  This was causing failed pods to "hang" since we were waiting for kubernetes to start them.

Once a pod is no longer running we then check success? and get the output, so by marking these failed pods as not running we allow it to go through the normal process for a completed pod.

lib/floe/workflow/runner/kubernetes.rb
agrare: Yeah trying to avoid more_core if possible, we only use it for one thing right now and we could refactor that and drop the dependency on that and activesupport all together.

kbrock:
The workflow will probably want to display a reasonable `cause`.
Not sure how we can expose this unless we tweak `output` to put in this error?

agrare:
@kbrock I've updated output to return a properly formatted error with `"Error"` and `"Cause"` keys

Here's an example of a failed workflow's output
```
{"Error"=>"ErrImagePull", "Cause"=>"rpc error: code = Unknown desc = failed to pull and unpack image \"docker.io/agrare/hello-worl:latest\": failed to resolve reference \"docker.io/agrare/hello-worl:latest\": pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed"}
```

And I can add Fail state matchers:
```
    "NextState": {
      "Type": "Task",
      "Resource": "docker://docker.io/agrare/hello-worl:latest",
      "Secrets": ["vmdb:aaa-bbb-ccc"],
      "End": true,
      "Catch": [
        {
          "ErrorEquals": [ "ErrImagePull", "ImagePullBackOff" ],
          "Next": "ImageFailState"
        }
      ]
    },

    "ImageFailState": {
      "Type": "Fail",
      "Error": "ImageFailStateError",
      "Cause": "ImageFailed"
    }
```

```
Running state: [ImageFailState]
{"Error"=>"ImageFailStateError", "Cause"=>"ImageFailed"}
```

kbrock:
This is looking great

Did you want to try and find an `Error` from https://docs.aws.amazon.com/step-functions/latest/dg/concepts-error-handling.html or are you thinking custom `Error` is good for now?

agrare:
I didn't want to just use the generic "States.TaskFailed" or "States.Runtime" since that is pretty broad and wouldn't allow people to handle these errors specifically, also we're good to define our own errors as long as they don't collide with the `States.` namespace

> States can report errors with other names. However, these error names can't begin with the States. prefix.

kbrock:
Thought `Cause` has the purpose of helping people understand what is wrong.

agrare:
> Thought Cause has the purpose of helping people understand what is wrong.

It does, Error is equivalent to the exception class and Cause is the message.  Did I screw that up somewhere? https://github.com/ManageIQ/floe/pull/135/files#diff-15bec43924080ff9b4e1850b9f3e3864476abea254de8b717a98816f856cb5a0R89

PR #140, opened 2023-11-06
Refactor the Podman runner to be a Docker subclass

agrare:
These two are so close that it seems like only a few overrides will allow us to remove a good amount of duplication


kbrock:
wow.
I see `run_continer` on the chopping block next (after this merge)

agrare:
>  I see run_continer on the chopping block next (after this merge)

100% just need to extract the params to a method then we can drop that one as well

PR #144, opened 2023-11-21
Remove more core dependency from ReferencePath

agrare:
TODO:
- [x] handle arrays

### DIFF
--- a/lib/floe/workflow/reference_path.rb
+++ b/lib/floe/workflow/reference_path.rb
@@ -10,9 +10,6 @@ module Floe
       end
 
       def initialize(*)
-        require "more_core_extensions/core_ext/hash/nested"
-        require "more_core_extensions/core_ext/array/nested"
-
         super
 
         raise Floe::InvalidWorkflowError, "Invalid Reference Path" if payload.match?(/@|,|:|\?/)
@@ -34,7 +31,12 @@ module Floe
         if path.empty?
           result.merge!(value)
         else
-          result.store_path(path, value)
+          child = result
+          path[..-2].each do |key|
+            child[key] = {} if child[key].nil?
+            child = child[key]
+          end
+          child[path.last] = value
         end
 
         result
--- a/lib/floe/workflow/reference_path.rb
+++ b/lib/floe/workflow/reference_path.rb
@@ -4,6 +4,10 @@ module Floe
   class Workflow
     class ReferencePath < Path
       class << self
+        def get(payload, context)
+          new(payload).get(context)
+        end
+
         def set(payload, context, value)
           new(payload).set(context, value)
         end
@@ -22,11 +26,17 @@ module Floe
                         .compact
       end
 
+      def get(context)
+        return context if path.empty?
+
+        context.dig(*path)
+      end
+
       def set(context, value)
         result = context.dup
 
         # If the payload is '$' then merge the value into the context
-        # otherwise use store path to set the value to a sub-key
+        # otherwise store the value under the path
         #
         # TODO: how to handle non-hash values, raise error if path=$ and value not a hash?
         if path.empty?
--- a/lib/floe/workflow/reference_path.rb
+++ b/lib/floe/workflow/reference_path.rb
@@ -32,12 +32,16 @@ module Floe
         if path.empty?
           result.merge!(value)
         else
-          child = result
-          path[..-2].each do |key|
+          child    = result
+          keys     = path.dup
+          last_key = keys.pop
+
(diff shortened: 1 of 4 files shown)


kbrock:
I also want more_core_extensions out since this is the only place where we use it.

I was concerned about accessing strings, hashes, or arrays.
It seemed like json supports all 3. (🍅 please)
Does this new code handle all 3?


Glad you moved the `@path` code into the initializer. I am wondering why we need the `map` stuff. I think I just don't understand it.

spec/workflow/reference_path_spec.rb
```diff
@@ -1,7 +1,78 @@
...
+      let(:input)   { {"hello" => "world"} }
+
+      it "returns the input" do
+        expect(subject.get(input)).to eq(input)
+      end
+    end
+
+    context "with an array dereference" do
+      let(:payload) { "$['store'][0]['book']" }
+      let(:input)   { {"store" => [{"book" => "ASL For Dummies"}]} }
```
kbrock, line 26: 0 seems like an easy case.
For some reason, I want to see 1 working

```suggestion
      let(:payload) { "$['store'][1]['book']" }
      let(:input)   { {"store" => [{},{"book" => "ASL For Dummies"}]} }
```

kbrock:
can we test:

1. The top level object is an array. (we only currently deal with hashes, and I think array is valid. To punt, we can pass in `$` as the path.
2. Is there a way to append to an array? (per the spec)

spec/workflow/reference_path_spec.rb
agrare: :+1: done (still works)

agrare:
> The top level object is an array. 

Absolutely, done.

> Is there a way to append to an array? (per the spec)

That is a great question I don't know

agrare:
> > Is there a way to append to an array? (per the spec)
>
> That is a great question I don't know

So it looks like `[-1:]` would mean the last element of an array, but the spec for reference paths specifically says that: `The operators "@", ",", ":", and "?" are not supported` so that wouldn't be valid.  I was able to easily set e.g. `$.[1].title` to create a second element in an array that only had one before though which seems like a valid case.

lib/floe/workflow/reference_path.rb
```diff
@@ -4,37 +4,60 @@ module Floe
...
 
         raise Floe::InvalidWorkflowError, "Invalid Reference Path" if payload.match?(/@|,|:|\?/)
+        @path = JsonPath.new(payload)
+                        .path[1..]
+                        .map { |v| v.match(/\[(?<name>.+)\]/)["name"] }
+                        .map { |v| v[0] == "'" ? v.delete("'") : v.to_i }
+                        .compact
+      end
+
+      def get(context)
```
Fryguy, line 29: I would have expected get to just be .dig?

lib/floe/workflow/reference_path.rb
```diff
@@ -4,37 +4,60 @@ module Floe
...
         result = context.dup
 
-        path = JsonPath.new(payload)
-                       .path[1..]
-                       .map { |v| v.match(/\[(?<name>.+)\]/)["name"] }
-                       .map { |v| v[0] == "'" ? v.delete("'") : v.to_i }
-                       .compact
-
         # If the payload is '$' then merge the value into the context
         # otherwise use store path to set the value to a sub-key
```
Fryguy, line 45: Minor but this comment refers to store path (as text so maybe it's ok)

Fryguy:
High level is the goal just to drop more core?  If so I'm curious why?

agrare:
Just trying to cut down on dependencies particularly because more_core brings in activesupport and trying to keep this "clean"

kbrock:
If you ignore the development gems in `Gemfile`, dropping `more_core_extensions` drops quite a few gems:

- more_core_extensions
  - active_support
    - base64
    - bigdecimal
    - concurrent-ruby
    - connection_pool
    - drb
      - ruby2_keywords
    - i18n
    - minitest
    - mutex_m
    - tzinfo
  - sync

---

ASIDE: kubeclient has even more dependencies, but not trivial to remove:

- kubeclient
  - http
    - addressable
      - public_suffix
    - http-cookie
      - domain_name
    - http-form_data
    - llhttp-ffi
      - ffi-compiler
        - ffi
  - jsonpath
  - recursive-open-struct
  - rest-client
    - http-accept
    - (http-cookie)
    - mime-types
      - mime-types-data
    - netrc

kbrock:
bump @Fryguy - you ok with this?

PR #146, opened 2023-12-06
[WIP] introduce container that echos the world

kbrock:
I promised @agrare I will post this
Not sure it is still needed

waiting on more examples

### DIFF
--- a/examples/echo-world/Dockerfile
+++ b/examples/echo-world/Dockerfile
@@ -1,0 +1,7 @@
+FROM docker.io/library/alpine as base
+
+WORKDIR /usr/src/app
+COPY echo-world.sh .
+CMD ["/usr/src/app/echo-world.sh"]
+
+
--- a/examples/echo-world/echo-world.sh
+++ b/examples/echo-world/echo-world.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env sh
+
+# docker run -t -i --rm -e STATUS=1 -e ERROR=error -e MESSAGE=abc kbrock/echo-world
+
+[ -n "${ERROR}" ]   && echo ${ERROR} >&2
+[ -n "${MESSAGE}" ] && echo ${MESSAGE}
+exit ${STATUS:-0}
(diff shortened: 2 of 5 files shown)


agrare:
So this could be used by the `examples/set-credentials.asl` from https://github.com/ManageIQ/floe/pull/145/files#diff-c6a697638a1e71ae8f717ce34601786d4c29f902b87e85e6a7449d5b17deb1ab

I do like having the docker images used by examples with the asl files :+1:

examples/echo-world/build.sh
```diff
@@ -0,0 +1,41 @@
...
+fi
+
+# may want to point to file's directory not pwd
+# APP=${APP:-${PWD##*/}}
+APP=${APP:-echo-world}
+
+echo "building '$APP' VERSION '${VERSION}'"
+
+# send to docker (no platform?)
+docker buildx build . --load                                    --tag kbrock/${APP}:${VERSION} --tag kbrock/${APP}:latest
```
agrare, line 34: My only issue with this is that `docker buildx` isn't available on most linux distributions' package repositories.  I can't `apt install docker` and run `docker buildx` for example.

It looks like buildah has added support for multiarch builds over in podman https://github.com/containers/buildah/issues/1590

Fryguy: This is how we do it for pods: https://github.com/ManageIQ/manageiq-pods/blob/master/bin/build#L57-L99.

For the most part docker build and podman build are basically the same, and if needed we can call out specific podman commands, but even the ones there in manageiq-pods wouldn't be needed here.

Fryguy: So, yeah, I agree with @agrare in that we don't need docker buildx.  Additionally I don't think we need the DOCKER_HOST (at least not in the script - my machine works directly without it and I think most other's machines do too)

The question is really whether we need multiarch (for which docker manifest is more than enough), or if linux/amd64 is enough, and whether or not we need the VERSION stuff.

agrare: We should be able to use what we have https://github.com/ManageIQ/workflows-examples/pull/18

Fryguy: agreed, I'd be +1 for an extra container in the workflows-examples repo

kbrock:
WIP: post example container elsewhere

kbrock:
closing deferring to other examples

PR #147, opened 2023-12-06
extra cli options

kbrock:
- Added logical shortcuts for parameters. (NOTE: docker runner changed from `-d` to `-r`)
- Added `--docker`, `--podman`, and `--kubernetes` shortcut

Before
=====

```
Usage: floe [options] workflow input [workflow2 input2]

v0.7.0

Options:
  -w, --workflow=<s>                  Path to your workflow json (legacy)
  -i, --input=<s>                     JSON payload to input to the workflow (legacy)
  -c, --credentials=<s>               JSON payload with credentials
  -r, --credentials-file=<s>          Path to a file with credentials
  -d, --docker-runner=<s>             Type of runner for docker images (default: docker)
  -o, --docker-runner-options=<s+>    Options to pass to the runner
  -v, --version                       Print version and exit
  -h, --help                          Show this message
```

After
======

```
Usage: floe [options] workflow input [workflow2 input2]

v0.7.0

Options:
  -w, --workflow=<s>                  Path to your workflow json (legacy)
  -i, --input=<s>                     JSON payload to input to the workflow (legacy)
  -c, --credentials=<s>               JSON payload with credentials
  -e, --credentials-file=<s>          Path to a file with credentials
  -r, --docker-runner=<s>             Type of runner for docker images
  -o, --docker-runner-options=<s+>    Options to pass to the runner
  -d, --docker                        Use docker to run images     (short for --docker_runner=docker)
  -p, --podman                        Use podman to run images     (short for --docker_runner=podman)
  -k, --kubernetes                    Use kubernetes to run images (short for --docker_runner=kubernetes)
  -v, --version                       Print version and exit
  -h, --help                          Show this message
```

### DIFF
--- a/exe/floe
+++ b/exe/floe
@@ -12,8 +12,12 @@ opts = Optimist.options do
   opt :input, "JSON payload to input to the workflow (legacy)", :type => :string
   opt :credentials, "JSON payload with credentials",            :type => :string
   opt :credentials_file, "Path to a file with credentials",     :type => :string
-  opt :docker_runner, "Type of runner for docker images",       :default => "docker"
-  opt :docker_runner_options, "Options to pass to the runner",  :type => :strings
+  opt :docker_runner, "Type of runner for docker images",       :type => :string, :short => 'r'
+  opt :docker_runner_options, "Options to pass to the runner",  :type => :strings, :short => 'o'
+
+  opt :docker,     "Use docker to run images     (short for --docker_runner=docker)",     :type => :boolean
+  opt :podman,     "Use podman to run images     (short for --docker_runner=podman)",     :type => :boolean
+  opt :kubernetes, "Use kubernetes to run images (short for --docker_runner=kubernetes)", :type => :boolean
 end
 
 Optimist.die(:docker_runner, "must be one of #{Floe::Workflow::Runner::TYPES.join(", ")}") unless Floe::Workflow::Runner::TYPES.include?(opts[:docker_runner])
@@ -21,6 +25,11 @@ Optimist.die(:docker_runner, "must be one of #{Floe::Workflow::Runner::TYPES.joi
 # legacy support for --workflow
 args = ARGV.empty? ? [opts[:workflow], opts[:input]] : ARGV
 Optimist.die(:workflow, "must be specified") if args.empty?
+
+# shortcut support
+opts[:docker_runner] ||= "docker" if opts[:docker]
+opts[:docker_runner] ||= "podman" if opts[:podman]
+opts[:docker_runner] ||= "kubernetes" if opts[:kubernetes]
 
 require "logger"
 Floe.logger = Logger.new($stdout)


kbrock:
update:

- added capitals
- added phrase shortcut

kbrock:
Todo:

- move input, creds to argv 
- don't use input, credentials from args

Prep for multiple workflows

kbrock:
update:
- removed argv changes (pulling into a separate PR)

exe/floe
```diff
@@ -6,13 +6,21 @@ require "optimist"
...
+  opt :docker_runner_options, "Options to pass to the runner",           :type => :strings, :short => 'o'
+  opt :docker,                "Use docker to run images  (shortcut)",    :type => :boolean, :short => 'd'
+  opt :podman,                "Use podman to run images  (shortcut)",    :type => :boolean, :short => 'p'
+  opt :kubernetes,            "Use kubernetes to run images (shortcut)", :type => :boolean, :short => 'k'
 end
 
+# alternative options
+opts[:docker_runner]="docker" if opts[:docker]
+opts[:docker_runner]="podman" if opts[:podman]
+opts[:docker_runner]="kubernetes" if opts[:kubernetes]
```
agrare, line 22:
```suggestion
opts[:docker_runner] = "docker" if opts[:docker]
opts[:docker_runner] = "podman" if opts[:podman]
opts[:docker_runner] = "kubernetes" if opts[:kubernetes]
```

agrare: Do we want to handle the case where they specify both the short and the long option differently?  e.g. `floe --podman --docker-runner=kubernetes`
I'd expect either an error, or the long-option to take precedence where this looks like the short-hand will take precedence.

kbrock: I would prefer to just drop the long form

Personally, `--docker` is much easier than `--docker_runner=docker`. And it basically needs to be specified every time

Currently, the long form is defaulted so it will always be set.
I'll change the way it handles defaults and have long form override

kbrock:
update:
- moved factory into a hash. used that hash for shortcuts, default, and factory
- changed shortcut verbiage from #149 

ok, I used the TYPES value to drive the shortcuts. is this too cryptic?

kbrock:
update:
- lost `usage` in rebases, added back
- updated git commit comment to reflect new `--help` output

kbrock:
ugh. sorry, got confused, usage gets changed in the other pr. (that is why it was missing...)

update:
- removed usage string

exe/floe
```diff
@@ -4,29 +4,37 @@
...
+  opt :docker_runner_options, "Options to pass to the runner", :type => :strings, :short => 'o'
+
+  supported_docker_runners.each_key do |runner|
+    opt runner.to_sym, "Use #{runner} to run images (short for --docker_runner=#{runner})", :type => :boolean
+  end
 end
 
-Optimist.die(:docker_runner, "must be one of #{Floe::Workflow::Runner::TYPES.join(", ")}") unless Floe::Workflow::Runner::TYPES.include?(opts[:docker_runner])
+# if they specified a docker_runner, and it is unknown, complain
+Optimist.die(:docker_runner, "must be one of #{supported_docker_runners.join(", ")}") if opts[:docker_runner] && !supported_docker_runners.key?(opts[:docker_runner])
```
kbrock, line 28: felt strange to have these defined in another file. but every one was then again listed in this file.

the main change here is only requiring the parameter if it was specified (before, it was defaulted so it was always present)

exe/floe
```diff
@@ -4,29 +4,37 @@
...
-                 Floe::Workflow::Runner::Podman
-               when "kubernetes"
-                 Floe::Workflow::Runner::Kubernetes
-               end
-
+# use docker_runner parameter, shortcut, or default
+runner_klass =
+  supported_docker_runners[opts[:docker_runner]] ||
+  supported_docker_runners.detect { |type, _klass| opts[type.to_sym] }&.last ||
+  supported_docker_runners.values.first
```
kbrock, line 37: I like the conversion from case to hash.

But the `detect` and code is not as clear

We could also do:

```ruby
opts[:docker_runner] ||= opts[:docker] && "docker" || opts[:kubernetes] && "kubernetes || opts[:podman] && "podman" || support_docker_runners.values.first
```

or what we originally had:

```ruby
opts[:docker_runner] ||= "docker" if opts[:docker]
opts[:docker_runner] ||= "podman" if opts[:podman]
opts[:docker_runner] ||= "kubernetes" if opts[:kubernetes]
opts[:docker_runner] ||= support_docker_runners.values.first
```

kbrock:
ok, code was too clever. I removed `hash_table.each` and just hardcoded the values.

exe/floe
```diff
@@ -4,29 +4,37 @@
...
   opt :input, "JSON payload to input to the workflow",         :default => '{}'
   opt :credentials, "JSON payload with credentials",           :default => "{}"
-  opt :docker_runner, "Type of runner for docker images",      :default => "docker"
-  opt :docker_runner_options, "Options to pass to the runner", :type => :strings
+  opt :docker_runner, "Type of runner for docker images",      :type => :string, :short => 'r'
+  opt :docker_runner_options, "Options to pass to the runner", :type => :strings, :short => 'o'
+
+  supported_docker_runners.each_key do |runner|
+    opt runner.to_sym, "Use #{runner} to run images (short for --docker_runner=#{runner})", :type => :boolean
+  end
```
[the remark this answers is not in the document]
kbrock: replaced

exe/floe
```diff
@@ -4,29 +4,39 @@
...
+
 opts = Optimist.options do
   version("v#{Floe::VERSION}\n")
+
   opt :workflow, "Path to your workflow json",                 :type => :string, :required => true
   opt :input, "JSON payload to input to the workflow",         :default => '{}'
   opt :credentials, "JSON payload with credentials",           :default => "{}"
-  opt :docker_runner, "Type of runner for docker images",      :default => "docker"
-  opt :docker_runner_options, "Options to pass to the runner", :type => :strings
+  opt :docker_runner, "Type of runner for docker images",      :type => :string, :short => 'r'
```
Fryguy, line 20: The upside to keep the "default" here is that the help text shows it's the default.  Compare:

```diff
- -d, --docker-runner=<s>             Type of runner for docker images (default: docker)
+ -d, --docker-runner=<s>             Type of runner for docker images
```

exe/floe
```diff
@@ -4,29 +4,39 @@
 require "floe"
 require "optimist"
 
+supported_docker_runners = {
+  "docker"     => Floe::Workflow::Runner::Docker,
+  "podman"     => Floe::Workflow::Runner::Podman,
+  "kubernetes" => Floe::Workflow::Runner::Kubernetes,
+  nil          => Floe::Workflow::Runner::Docker # default
+}
```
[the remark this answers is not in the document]
agrare: Don't we already have `Floe::Workflow::Runner::TYPES` for this purpose?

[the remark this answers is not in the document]
Fryguy: LOL yes - I totally forgot about that constant :)

exe/floe
kbrock: if docker-runner is defaulted, then we do not know if someone specifies docker with `--docker-runner=docker` (and we should ignore a `-k`) or if we should go with kubernetes because they did not specify a `docker-runner` and rather wanted

exe/floe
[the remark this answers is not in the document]
kbrock: I would like the factory to be the one that specifies the values that are used in the factory.

I brought those locally. But I could have moved that change into the Runner (maybe converting that into a hash)

exe/floe
kbrock: also tried:

```ruby
SUPPORTED_DOCKER_RUNNERS.each_key do |runner_name|
  opts[:docker_runner] ||= runner_name if opts[runner_name.to_sym]
end
```

Keep going back to the duplicate code version. It takes no braincells to read and is the same number of lines.

kbrock:
Updates:

- removed `TYPES` from runner.
- `DEFAULT` is a separate variable
- `Optimist.die` is moved and simplified

I like it how it is.

But there are a few possible changes:

- If we want to keep `Runner::TYPES`, then we can introduce `set_runner(name="docker", options={})` and move the factory into there.
- If we want a default for `docker-runner`, then we will allow short flags to override longer flags.
- I keep wanting to change the code to have the options generated from `TYPES`, but find the code more confusing and not any shorter. So I keep tearing it back out.

kbrock:
update:

- extracted factory / constant definition into #152
- only introduce aliases.
- still changes the short form of credentials file, docker runner and docker runner options (to `-e`, `-r` and `-o`)

Outstanding

- I did like the `docker_runner, :default => "docker"`, but we can't keep that and have long form override short form. So I dropped the default. Can have last one wins for short form (instead of first one wins)

agrare (approved):

PR #148, opened 2023-12-06
allow non quoted json keys for input

kbrock:
Before
======

```
bundle exec exe/floe --input='{"foo": 2}' examples/run-json.asl
```

After
=====

```
bundle exec exe/floe --input='{foo: 2}' examples/run-json.asl
```


agrare:
@kbrock isn't `{foo: 2}` invalid json?

Fryguy:
I'm not sure about this, because it makes it appear like invalid json is acceptable.

kbrock:
punting on this idea

PR #152, opened 2023-12-20
Create docker runners factory and add scheme

kbrock:
Defining the runners in a single place

- extracted from #147
- dependent upon this PR: https://github.com/ManageIQ/manageiq-providers-workflows/pull/63

Before
======

- types specified in came from `Floe::Workflow::Runner::TYPES` and `exe/floe`
- class names specified in `exe/floe` and `manageiq-providers-workflow`
- default specified in `Floe::Workflow::Runner::TYPES` and `exe/floe`
- users of floe need to know internal class names

After
=====

- types and class names only specified in `Floe::Workflow::Runner`
- use `Floe.set_runner` to specify the name of the runner
- deprecated old `docker_runner=` interface
- support multiple schemes

### DIFF
--- a/lib/floe.rb
+++ b/lib/floe.rb
@@ -59,12 +59,13 @@ module Floe
   # Set the runner to use
   #
   # @example
-  #   Floe.set_runner "kubernetes"
-  #   Floe.set_runner Floe::Workflow::Runner::Kubernetes.new
+  #   Floe.set_runner "docker", kubernetes", {}
+  #   Floe.set_runner "docker", Floe::Workflow::Runner::Kubernetes.new({})
   #
+  # @param scheme [String] scheme                           Protocol to register (e.g.: docker)
   # @param name_or_instance [String|Floe::Workflow::Runner] Name of runner to use for docker (e.g.: docker)
   # @param options [Hash]                                   Options for constructor of the runner (optional)
-  def self.set_runner(name, options = {})
-    Floe::Workflow::Runner.set_runner(name, options)
+  def self.set_runner(scheme, name_or_instance, options = {})
+    Floe::Workflow::Runner.set_runner(scheme, name_or_instance, options)
   end
 end
--- a/lib/floe.rb
+++ b/lib/floe.rb
@@ -45,7 +45,26 @@ module Floe
     @logger ||= NullLogger.new
   end
 
+  # Set the logger to use
+  #
+  # @example
+  #   require "logger"
+  #   Floe.logger = Logger.new($stdout)
+  #
+  # @param logger [Logger] logger to use for logging actions
   def self.logger=(logger)
     @logger = logger
+  end
+
+  # Set the runner to use
+  #
+  # @example
+  #   Floe.set_runner "kubernetes"
+  #   Floe.set_runner Floe::Workflow::Runner::Kubernetes.new
+  #
+  # @param name_or_instance [String|Floe::Workflow::Runner] Name of runner to use for docker (e.g.: docker)
+  # @param options [Hash]                                   Options for constructor of the runner (optional)
+  def self.set_runner(name, options = {})
+    Floe::Workflow::Runner.set_runner(name, options)
   end
 end
--- a/lib/floe/workflow/runner.rb
+++ b/lib/floe/workflow/runner.rb
@@ -5,17 +5,36 @@ module Floe
     class Runner
       include Logging
 
-      TYPES         = %w[docker podman kubernetes].freeze
       OUTPUT_MARKER = "__FLOE_OUTPUT__\n"
 
       def initialize(_options = {})
       end
 
       class << self
-        attr_writer :docker_runner
+        # deprecated -- use Floe.set_runner instead
+        def docker_runner=(value)
+          set_runner(value)
+        end
+
+        # see Floe.set_runner
+        def set_runner(name_or_instance, options = {})
+          @docker_runner =
+            case name_or_instance
+            when "docker", nil
+              Floe::Workflow::Runner::Docker.new(options)
+            when "podman"
+              Floe::Workflow::Runner::Podman.new(options)
+            when "kubernetes"
+              Floe::Workflow::Runner::Kubernetes.new(options)
+            when Floe::Workflow::Runner
+              name_or_instance
+            else
+              raise ArgumentError, "docker runner must be one of: docker, podman, kubernetes"
+            end
+        end
 
         def docker_runner
-          @docker_runner ||= Floe::Workflow::Runner::Docker.new
+          @docker_runner || set_runner("docker")
         end
 
         def for_resource(resource)
--- a/lib/floe/workflow/runner.rb
+++ b/lib/floe/workflow/runner.rb
@@ -10,15 +10,16 @@ module Floe
       def initialize(_options = {})
       end
 
+      @runners = {}
       class << self
         # deprecated -- use Floe.set_runner instead
         def docker_runner=(value)
-          set_runner(value)
+          set_runner("docker", value)
         end
 
         # see Floe.set_runner
-        def set_runner(name_or_instance, options = {})
-          @docker_runner =
+        def set_runner(scheme, name_or_instance, options = {})
+          @runners[scheme] =
             case name_or_instance
             when "docker", nil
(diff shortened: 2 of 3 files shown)


kbrock:
update:
- fixed ArgumentError (when you specify an invalid docker runner name

kbrock:
Complexities that I can remove if you want:

1. can remove supporting multiple schemas
2. can remove supporting register with or without schema

kbrock:
update:

- split into 2 commits (1 for factory, 1 for schemas)
- ensured it supports passing in workflow instance or strings

update:

- fixed bad rebase

lib/floe/workflow/runner.rb
```diff
@@ -5,29 +5,55 @@ class Workflow
...
+        #
+        # example
+        #   register_runner "kubernetes", {}
+        #   register_runner nil # use default
+        #   register_runner Floe::Workflow::Runner::Kubernetes.new
+        #
+        #   # specify with a schema
+        #   register_runner "docker", nil # use default runner (docker)
+        #   register_runner "docker", "kubernetes", {}
+        def register_runner(schema = "docker", name_or_instance = {}, options = {})
```
Fryguy, line 29: Register reads funny to me.  When i see a register method, I think of it like adding something to a registry.  So, in plugin-style libraries, I think of the register method being called for each of the various plugins to put each of them into the collective registry.

Instead this method is just picking a runner based on a symbol/string, which, to me, is more like what I'd expect from `docker_runner=` itself.  Perhaps we can just use that method directly?  Since we are < 1.0 release, I think it's ok to have a breaking change like that.

Fryguy: Or perhaps `select_docker_runner` or `set_docker_runner`?

kbrock: I would prefer if floe knew about the classes associated with the runners rather than workflow or `exe/floe`.

But if we think the caller should know about these things, then we can go in that direction. I'll see if I still have the branch around that had that.

kbrock:
update:
- extracted schema changes into a separate PR

kbrock:
update:
- updated version number to 0.8.0
- dropped docker_runner=
- added entries to readme and changelog

lib/floe/workflow/runner.rb
kbrock: I renamed this method. think it is what you want.

lib/floe.rb
```diff
@@ -45,7 +45,27 @@ def self.logger
...
   def self.logger=(logger)
     @logger = logger
   end
+
+  # Set the runner to use
+  #
+  # example
+  #
+  #   Floe.set_runner "kubernetes"
+  #   Floe.set_runner Floe::Workflow::Runner::Kubernetes.new
```
Fryguy, line 64:
```suggestion
  # @example
  #   Floe.set_runner "kubernetes"
  #   Floe.set_runner Floe::Workflow::Runner::Kubernetes.new
```

lib/floe.rb
```diff
@@ -45,7 +45,27 @@ def self.logger
...
+  #
+  # @param name_or_instance [String|Floe::Workflow::Runner] name of runner to use for docker (default: docker)
```
Fryguy, line 66:
```suggestion
  # @param name_or_instance [String|Floe::Workflow::Runner] Name of runner to use for docker (default: docker)
```

lib/floe.rb
```diff
@@ -45,7 +45,27 @@ def self.logger
     @logger ||= NullLogger.new
   end
 
+  # Set the logger to use
+  #
+  # example
+  #
+  #   require "logger"
+  #   Floe.logger = Logger.new($stdout)
+  # @param logger [Logger] logger to use for debugging
```
Fryguy, line 54:
```suggestion
  # @example
  #   require "logger"
  #   Floe.logger = Logger.new($stdout)
  #
  # @param logger [Logger] logger to use for debugging
```

Fryguy: I'm not sure we should say "for debugging", it's also just for regular logging of actions.

kbrock:
update:
- rebased
- example tags
- uppercased yarddoc params (Name)
- changed logger comment to "logging actions"

lib/floe/workflow/runner.rb
```diff
@@ -5,17 +5,31 @@ class Workflow
...
       OUTPUT_MARKER = "__FLOE_OUTPUT__\n"
 
       def initialize(_options = {})
       end
 
       class << self
-        attr_writer :docker_runner
+        # see Floe.set_runner
+        def set_runner(name_or_instance = nil, options = {})
+          @docker_runner =
```
agrare, line 16: @kbrock I think leaving the `docker` part out of this is confusing, because we expect to have other types of runners (e.g. lambda, miq/internal, ansible).  I agree it is confusing that there is a docker-docker runner but I think we still need to call out the fact that we're setting a runner for a particular scheme (in this case `docker://`).

CHANGELOG.md
```diff
@@ -3,13 +3,16 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Breaking Change
+- `Floe::Workflow::Runner.docker_runner=` is now `Floe.set_runner` ([#152](https://github.com/ManageIQ/floe/pull/152))
```
agrare, line 7: Could we add an alias so we don't have to do a major version bump, just deprecate the old one and drop it when we do a major bump?

lib/floe/workflow/runner.rb
kbrock: that is available at https://github.com/ManageIQ/floe/pull/153

kbrock:
update:
- merged #153 into this one.

I had thought that pulling out the addition of a `scheme` would simplify things, but instead it has only added naming questions and confusion.

agrare (approved):

lib/floe.rb
```diff
@@ -45,7 +45,27 @@ def self.logger
...
+  #
+  # @param logger [Logger] logger to use for logging actions
   def self.logger=(logger)
     @logger = logger
   end
+
+  # Set the runner to use
+  #
+  # @example
+  #   Floe.set_runner "docker", kubernetes", {}
```
Fryguy, line 62: super tiny typo here

```suggestion
  #   Floe.set_runner "docker", "kubernetes", {}
```

lib/floe/workflow/runner.rb
```diff
@@ -5,29 +5,42 @@ class Workflow
...
       class << self
-        attr_writer :docker_runner
+        # deprecated -- use Floe.set_runner instead
+        def docker_runner=(value)
+          set_runner("docker", value)
+        end
 
-        def docker_runner
-          @docker_runner ||= Floe::Workflow::Runner::Docker.new
+        # see Floe.set_runner
```
Fryguy, line 20: yard supports see references directly ([docs](https://rubydoc.info/gems/yard/file/docs/Tags.md#see))

```suggestion
        # @see Floe.set_runner
```

lib/floe/workflow/runner.rb
```diff
@@ -5,29 +5,42 @@ class Workflow
...
-      TYPES         = %w[docker podman kubernetes].freeze
       OUTPUT_MARKER = "__FLOE_OUTPUT__\n"
 
       def initialize(_options = {})
       end
 
+      @runners = {}
       class << self
-        attr_writer :docker_runner
+        # deprecated -- use Floe.set_runner instead
```
Fryguy, line 15: yard has a deprecated tag we can use ([docs](https://rubydoc.info/gems/yard/file/docs/Tags.md#deprecated))

```suggestion
        # @deprecated Use Floe.set_runner instead
```

lib/floe/workflow/runner.rb
```diff
@@ -5,29 +5,42 @@ class Workflow
     class Runner
       include Logging
 
-      TYPES         = %w[docker podman kubernetes].freeze
       OUTPUT_MARKER = "__FLOE_OUTPUT__\n"
 
       def initialize(_options = {})
       end
 
+      @runners = {}
```
Fryguy, line 13: Do we have to worry about threaded access to this?  I could see floe being used in a multi-threaded scenario in the future.

PR #153, opened 2024-01-03
allow multiple runners/schemes

kbrock:
built upon:
- [ ] #152 (first two commits)

this adds scheme to `set_docker_runner`

### DIFF
--- a/lib/floe/workflow/runner.rb
+++ b/lib/floe/workflow/runner.rb
@@ -5,17 +5,36 @@ module Floe
     class Runner
       include Logging
 
-      TYPES         = %w[docker podman kubernetes].freeze
       OUTPUT_MARKER = "__FLOE_OUTPUT__\n"
 
       def initialize(_options = {})
       end
 
       class << self
-        attr_writer :docker_runner
+        # deprecated -- use Floe.set_runner instead
+        def docker_runner=(value)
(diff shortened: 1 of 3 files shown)


lib/floe/workflow/runner.rb
```diff
@@ -5,29 +5,50 @@ class Workflow
...
+              Floe::Workflow::Runner::Kubernetes.new(options)
+            when Floe::Workflow::Runner
+              name_or_instance
+            else
+              raise ArgumentError, "docker runner must be one of: docker, podman, kubernetes"
+            end
+        end
+
+        def docker_runner(scheme)
+          set_runner if @runners.empty?
```
Fryguy, line 43: Shouldn't the scheme be passed along?

```suggestion
        def docker_runner(scheme)
          set_runner(schema) if @runners.empty?
```

kbrock: This sets the default runner.

kbrock:
update:
- scheme is no longer optional
- no longer have default values for scheme/instance name.
- rename some cases of schema to scheme

update:
- fix `exe/floe` to pass scheme along
- fix refactor bug where scheme was not created correctly

kbrock:
merged into #152

PR #156, opened 2024-01-17
Ensure that wf.credentials is not-nil

agrare:
Now that we are allowing assignment to Credentials we have to ensure it is at least an empty hash.

We already were defaulting to `{}` if not specified but it was possible for the caller to pass an explicit `nil` in.


agrare:
I released v0.8.0 since there were other enhancements in master but I'm going to create a v0.7.z branch with just this fix.

agrare:
Released v0.7.1 with this change

PR #158, opened 2024-01-18
step_nonblock_wait is no longer used

kbrock:
If we need this method, then don't delete it. I just wanted to highlight that the code was not being called from here or provider-workflow

We are running wait/step_nonblocking_ready? on a list of workflows and no longer on a single workflow

After merging https://github.com/ManageIQ/floe/pull/157, I double checked that all the methods were being called...

Followup / or what this PR should be about:

I do feel that the main loop in `exe/floe` should be moved into `Workflow.` The version of `wait` isn't quite right and I like the version of `exe/floe` better.

### DIFF
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -74,10 +74,6 @@ module Floe
       current_state.run_nonblock!
     end
 
-    def step_nonblock_wait(timeout: nil)
-      current_state.wait(:timeout => timeout)
-    end
-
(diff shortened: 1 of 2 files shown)


agrare:
I think this is still a useful method to have as part of the `Floe::Workflow` interface even if we don't use it in `exe/floe` right now.  Plus eventually `step_nonblock_wait` could interrupt once the state is ready not just sleep

PR #159, opened 2024-02-14
Don't set next_state until finish for Wait state

agrare:
By setting the next_state in Wait#start we were bypassing the Wait#finish step because the next_state was already set.  This caused missing entries in the context for the state completion.

Before:
```
$ bundle exec exe/floe --podman --workflow examples/wait.asl --input '{"foo": 1}'
I, [2024-02-14T11:19:44.894906 #87255]  INFO -- : checking 1 workflows...
I, [2024-02-14T11:19:44.895029 #87255]  INFO -- : Running state: [FirstState] with input [{"foo"=>1}]...
D, [2024-02-14T11:19:44.895117 #87255] DEBUG -- : Running podman run --detach -e foo\=1 --name floe-sleep-4f985a02 localhost/sleep:latest
I, [2024-02-14T11:19:59.828972 #87255]  INFO -- : Running state: [FirstState] with input [{"foo"=>1}]...Complete - next state: [WaitState] output: [{"foo"=>1}]
I, [2024-02-14T11:20:00.035521 #87255]  INFO -- : Running state: [WaitState] with input [{"foo"=>1}]...
I, [2024-02-14T11:20:05.000492 #87255]  INFO -- : Running state: [SuccessState] with input [{"foo"=>1}]...
I, [2024-02-14T11:20:05.000658 #87255]  INFO -- : Running state: [SuccessState] with input [{"foo"=>1}]...Complete - next state: [] output: [{"foo"=>1}]
I, [2024-02-14T11:20:05.000713 #87255]  INFO -- : checking 1 workflows...Complete - 1 ready
{"foo"=>1}
```

Notice there is no `Running state: [WaitState] with input [{"foo"=>1}]...Complete` log line

### DIFF
--- a/lib/floe/workflow/states/wait.rb
+++ b/lib/floe/workflow/states/wait.rb
@@ -28,15 +28,19 @@ module Floe
 
         def start(input)
           super
-          input = input_path.value(context, input)
 
-          context.output     = output_path.value(context, input)
-          context.next_state = end? ? nil : @next
+          input          = input_path.value(context, input)
+          context.output = output_path.value(context, input)
 
           wait_until!(
             :seconds => seconds_path ? seconds_path.value(context, input).to_i : seconds,
             :time    => timestamp_path ? timestamp_path.value(context, input) : timestamp
           )
+        end
+
+        def finish
+          context.next_state = end? ? nil : @next
+          super
         end
 
         def running?
--- a/lib/floe/workflow/states/wait.rb
+++ b/lib/floe/workflow/states/wait.rb
@@ -38,11 +38,6 @@ module Floe
           )
         end
 
-        def finish
-          context.next_state = end? ? nil : @next
-          super
-        end
-
         def running?
           waiting?
         end
--- a/lib/floe/workflow/states/non_terminal_mixin.rb
+++ b/lib/floe/workflow/states/non_terminal_mixin.rb
@@ -4,6 +4,11 @@ module Floe
   class Workflow
     module States
       module NonTerminalMixin
+        def finish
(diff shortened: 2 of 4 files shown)


lib/floe/workflow/states/wait.rb
```diff
@@ -28,17 +28,21 @@ def initialize(workflow, name, payload)
...
+          context.output = output_path.value(context, input)
 
           wait_until!(
             :seconds => seconds_path ? seconds_path.value(context, input).to_i : seconds,
             :time    => timestamp_path ? timestamp_path.value(context, input) : timestamp
           )
         end
 
+        def finish
+          context.next_state = end? ? nil : @next
```
Fryguy, line 42: Wouldn't this be the same for all states?  I'm curious why wait is different.  I woiuld have expected that any state would describe the next state immediately (except for possibly choice)

agrare: Yeah true, not all states have Next but I could move this to the NonTerminalMixin

agrare:
@kbrock yeah exactly, Wait and Task are the only two currently which aren't `ready?` immediately, and Task already set next_state in `#finish`

kbrock:
Would it make sense to put finish into everything and put the next_state change into fail, pass, succeed, task?
(maybe choice?)

agrare:
> Would it make sense to put finish into everything and put the next_state change into fail, pass, and succeed even though they are ready immediately.

I think that's what I've done, every state that has `Next` sets next_state in finish now.  Task is a little different as you mentioned which is why I clear the next state in the failure case.

PR #161, opened 2024-03-22
Add ability to specify the API on the command line

kbrock:
Our current workflows have the api endpoint url stored in `Context["Execution"]["_manageiq_api_url"]`.

Before
======

There is no way to set the api url in exe/floe

After
=====

```
./exe/floe --context '{"Execution": {"_manageiq_api_url": "http://localhost:3000"}}' workflow.asl "{}"
```

followup:
- [ ] https://github.com/ManageIQ/manageiq-providers-workflows/pull/73

Comments
========

The `input` changes per workflow, so it makes sense to specify alongside the workflow in the ARGV.

The api endpoint url, on the other hand, is the same across all executions, so a single value make sense.

### DIFF
--- a/exe/floe
+++ b/exe/floe
@@ -10,6 +10,7 @@ opts = Optimist.options do
 
   opt :workflow, "Path to your workflow json (legacy)",         :type => :string
   opt :input, "JSON payload to input to the workflow (legacy)", :type => :string
+  opt :context, "JSON payload of the Context",                  :type => :string
   opt :credentials, "JSON payload with credentials",            :type => :string
   opt :credentials_file, "Path to a file with credentials",     :type => :string
   opt :docker_runner, "Type of runner for docker images",       :type => :string, :short => 'r'
@@ -49,7 +50,7 @@ credentials =
 
 workflows =
   args.each_slice(2).map do |workflow, input|
-    context = Floe::Workflow::Context.new(:input => input || opts[:input] || "{}")
+    context = Floe::Workflow::Context.new(opts[:context], :input => input || opts[:input] || "{}")
     Floe::Workflow.load(workflow, context, credentials)
   end
 
--- a/lib/floe/workflow/context.rb
+++ b/lib/floe/workflow/context.rb
@@ -3,19 +3,19 @@
 module Floe
   class Workflow
     class Context
+      # @param context [Json|Hash] (default, create another with input and execution params)
+      # @param input [Hash] (default: {})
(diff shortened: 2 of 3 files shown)


Fryguy (changes requested):
> This name feels very manageiq specific.

That's because it is manageiq specific.  Anything manageiq specific should not be in this repository.

I think a more generic way to handle this is to take the execution state as a parameter instead, similar to how we take input or credentials.

agrare:
@kbrock the API passing is something specific to manageiq, could you just pass the execution as part of the context input that exists already?

agrare:
> context input that exists already?

Looks like we only accept `--input` on the command-line and build a context from that, I think if we also had a `--context` option you could do `floe --context '{"Execution": {"_manageiq_api_url": "http://localhost:3000"}} ...'

kbrock:
update:
- setting context (allows a partial context)

kbrock:
🍏 , but code climate says this is more complex. Which it is.
But ensuring all these extra fields work.

Fryguy:
I take code climate complexity with a grain of salt 😅

exe/floe
```diff
@@ -10,6 +10,7 @@ opts = Optimist.options do
 
   opt :workflow, "Path to your workflow json (legacy)",         :type => :string
   opt :input, "JSON payload to input to the workflow (legacy)", :type => :string
+  opt :context, "Context",                                      :type => :string
```
agrare, line 13: Suggest `JSON payload of the context`

lib/floe/workflow/context.rb
```diff
@@ -3,19 +3,20 @@
...
-          "StateMachine" => {},
-          "Task"         => {}
-        }
+        @context = context || {}
+        self["Execution"] ||= {}
+        self["Execution"]["Input"] ||= input
+        self["State"] ||=        {}
+        self["StateHistory"] ||= []
+        self["StateMachine"] ||= {}
+        self["Task"] ||=         {}
```
agrare, line 19:
```suggestion
        self["Execution"]          ||= {}
        self["Execution"]["Input"] ||= input
        self["State"]              ||= {}
        self["StateHistory"]       ||= []
        self["StateMachine"]       ||= {}
        self["Task"]               ||= {}
```

agrare (approved):

agrare:
@kbrock can you just update the title and description to match what we ended up doing?

PR #162, opened 2024-03-27
Add simplecov

kbrock:
not tested but copied from manageiq

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,8 @@ jobs:
         - '3.0'
         - '3.1'
         - '3.2'
+    env:
+      CC_TEST_REPORTER_ID: "${{ secrets.CC_TEST_REPORTER_ID }}"
     steps:
     - uses: actions/checkout@v4
     - name: Set up Ruby
@@ -26,3 +28,7 @@ jobs:
         bundler-cache: true
     - name: Run tests
       run: bundle exec rake
+    - name: Report code coverage
+      if: "${{ github.ref == 'refs/heads/master' && matrix.ruby-version == '3.2' }}"
+      continue-on-error: true
+      uses: paambaati/codeclimate-action@v5
--- a/Gemfile
+++ b/Gemfile
@@ -7,9 +7,3 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 
 # Specify your gem's dependencies in floe.gemspec
 gemspec
-
-gem "manageiq-style"
-gem "rake", "~> 13.0"
-gem "rspec"
-gem "rubocop"
-gem "timecop"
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -12,4 +12,5 @@ gem "manageiq-style"
 gem "rake", "~> 13.0"
 gem "rspec"
 gem "rubocop"
+gem 'simplecov', require: false
```
Fryguy, line 15: this should be a dev dependency in the gemspec (in fact all of these should - not sure why they aren't)

spec/spec_helper.rb
```diff
@@ -12,6 +12,14 @@
 # the additional setup, and require it from the spec files that actually need
 # it.
 #
+
+if ENV['CI']
+  require 'simplecov'
+  SimpleCov.start do
+    add_filter "/spec/"
+  end
+end
```
Fryguy, line 21: > not tested but copied from manageiq

Probably better to copy from another gem as opposed to an application.

So, here it would just be (I looked at linux_admin as an example)

```suggestion
if ENV['CI']
  require 'simplecov'
  SimpleCov.start
end
```

Gemfile
Fryguy: Also simplecov needs a minimum for codeclimate - looking at another gem, it's

```
spec.add_development_dependency "simplecov", ">= 0.21.2"
```

.github/workflows/ci.yaml
```diff
@@ -26,3 +26,7 @@ jobs:
         bundler-cache: true
     - name: Run tests
       run: bundle exec rake
+    - name: Report code coverage
+      if: "${{ github.ref == 'refs/heads/master' && matrix.ruby-version == '3.2' }}"
+      continue-on-error: true
+      uses: paambaati/codeclimate-action@v5
```
Fryguy, line 32: This won't work without the secret.  Looking at another gem, add this above the steps

```
    env:
      CC_TEST_REPORTER_ID: "${{ secrets.CC_TEST_REPORTER_ID }}"
```

Example: https://github.com/ManageIQ/linux_admin/blob/746d84d93827ef930f31ee7611e93d56c8c83e84/.github/workflows/ci.yaml#L19-L20

spec/spec_helper.rb
kbrock: yea. saw no need to run simplecov on specs, since they always display 100%

kbrock:
update:
- moved gemfile dependencies
- fixed simplecov `require`
- added minimum version for simplecov
- added secret

update
- fixed gemspec type

spec/spec_helper.rb
Fryguy: I was certain I had a script for new repositories and ran it on here.  It should have included all this 😕

Fryguy (approved):

PR #163, opened 2024-03-27
Merging parse_output and parse_error

kbrock:
removing duplicate code

curious how code climate responds (think it will get worse)

Have run a preliminary run, but unsure the implications of `parse_error` changing from `output.nil?` to `output.blank?`.


kbrock:
adding `output.blank?` is probably a good thing.
But it says it is more complicated (6 => 9)
and since we aren't dinged for duplicate code, just punting for now

PR #164, opened 2024-03-28
fix rubocops

kbrock:


kbrock:
update:

- rebase

lib/floe/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Floe
-  VERSION = "0.9.0".freeze
+  VERSION = "0.9.0"
```
Fryguy, line 4: This one surprised me - I thought rubocop enforce freezing all constants 😕

lib/floe/workflow/reference_path.rb
```diff
@@ -19,11 +19,11 @@ def initialize(*)
         super
 
         raise Floe::InvalidWorkflowError, "Invalid Reference Path" if payload.match?(/@|,|:|\?/)
+
         @path = JsonPath.new(payload)
                         .path[1..]
                         .map { |v| v.match(/\[(?<name>.+)\]/)["name"] }
-                        .map { |v| v[0] == "'" ? v.delete("'") : v.to_i }
-                        .compact
+                        .filter_map { |v| v[0] == "'" ? v.delete("'") : v.to_i }
```
Fryguy, line 26: Nice! I didn't know filter_map was available on the versions of Ruby we supported (presuming test cases exercise this line)

lib/floe/version.rb
Fryguy: Oh duh, there's a frozen string literal at the top (I never liked those, and almost think that's the actual fix we should do here).

lib/floe/workflow/reference_path.rb
kbrock: just learned this one. `rubocop` does teach us good stuff. It is ruby>=2.7

Fryguy (approved):

PR #167, opened 2024-03-29
Drop non-standard Error/Cause fields.

kbrock:
We had introduced this to access the errors.
At that time exceptions were being thrown and it was tricky to determine failed runs.

Now output is properly populated with Error/Cause.

So we can use the standard `context.output` and no longer need these.

### DIFF
--- a/lib/floe/workflow/context.rb
+++ b/lib/floe/workflow/context.rb
@@ -30,6 +30,10 @@ module Floe
         started? && !ended?
       end
 
+      def failed?
+        output&.key?("Error") || false
+      end
+
       def ended?
         execution.key?("EndTime")
       end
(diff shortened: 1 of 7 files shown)


lib/floe/workflow/context.rb
```diff
@@ -67,7 +67,7 @@ def status
           "pending"
         elsif running?
           "running"
-        elsif state["Error"]
+        elsif output&.[]("Error")
```
agrare, line 70: @kbrock could we move this check to a `failure?` method like we have above for `started?` and `running?` ?

kbrock:
update:
- added `failed?`

agrare (approved):

PR #168, opened 2024-03-29
Set manageiq-style to at least 1.5.2

Fryguy:
I noticed locally that manageiq-style was being locked to 1.3.3, because rubocop was taking precedence and thus choosing a more liberal manageiq-style (i.e. 1.3.3 did not lock down the rubocop version).  Instead, here, we lock manageiq-style, which forces rubocop to the version we expect.

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -1,8 +1,6 @@
----
 extends: relaxed
-
 rules:
   indentation:
     indent-sequences: false
   line-length:
-    max: 120
+    max: 1000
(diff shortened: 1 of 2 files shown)


Fryguy:
> huh. so it has a mimimum ruby of 3.0

yeah that's annoying - I know why I did it - thinking I might downgrade or possibly remove that.

kbrock:
kicking

kbrock:
I added https://github.com/ManageIQ/manageiq-style/pull/49 to try and move this forward

agrare (approved):

PR #169, opened 2024-04-01
Add support for pluggable schemes

### LINKED ISSUE #45: Support pluggable resources
Having alternative runner that live outside of floe would be a cool feature.  For example, let's say in the ManageIQ application I wanted to run "builtin" methods, I would like to define, for example, `builtin://email`, or if I wanted to expose helpers for accessing data from our api, I would like to define `vmdb://api/providers`.

I believe this was initially implemented with a simple const_get, but I'd like to define these elsewhere and then "register" the resource type to a specific class, to avoid just loading any random constant.

Fryguy:
The idea behind this PR is to move the docker runner (renamed here to container runner for clarity) into a separate directory in a way to show what a plugin would look like and what the registration of schemes would look like.

For now, this PR:
- Move docker schema runner into a Floe::Runner::ContainerRunner namespace to treat it like a plugin
- Add registration of schema to runner instance (with support for procs)
- Change exe/floe to not have anything schema-specific and to call out to the plugins for options and post-option setup.

TODO:
- I want to add specs around the registration and the container runner's class methods for registration.  Not sure if it's needed for this PR or not.

Here's what the `--help` looks like after this:

```
$ be exe/floe --help
Usage: floe [options] workflow input [workflow2 input2]

v0.9.0

Options:
  -w, --workflow=<s>                     Path to your workflow json (legacy)
  -i, --input=<s>                        JSON payload to input to the workflow (legacy)
  -c, --context=<s>                      JSON payload of the Context
  -e, --credentials=<s>                  JSON payload with credentials
  -d, --credentials-file=<s>             Path to a file with credentials

Container runner options:
  -r, --container-runner=<s>             Type of runner for docker container images (docker, podman, or kubernetes)
  -o, --container-runner-options=<s+>    Options to pass to the container runner
  -k, --docker                           Use docker to run container images     (short for --container-runner=docker)
  -p, --podman                           Use podman to run container images     (short for --container-runner=podman)
  -u, --kubernetes                       Use kubernetes to run container images (short for --container-runner=kubernetes)

General options:
  -v, --version                          Print version and exit
  -h, --help                             Show this message
```

Some future thoughts:
- I actually wanted `Floe::ContainerRunner` directly, but didn't want to deal with the un-indenting of 1 level of namespace.  I still may do this anyway, because then this would mirror what a real "plugin" would look like allowing for things like `require 'floe/container_runner'` and living in a `floe-container_runner` gem.
- While the exe/floe hardcodes the reference to the container runner, I think it's cleaner to loop over all of the registered schemes, and then call cli_options/resolve_cli_options on each. However, I thought it would make this PR less clear and didn't do that for now.

@agrare @kbrock Please review.

Closes #45

### DIFF
--- a/lib/floe.rb
+++ b/lib/floe.rb
@@ -18,10 +18,6 @@ require_relative "floe/workflow/payload_template"
 require_relative "floe/workflow/reference_path"
 require_relative "floe/workflow/retrier"
 require_relative "floe/workflow/runner"
-require_relative "floe/workflow/runner/docker_mixin"
-require_relative "floe/workflow/runner/docker"
-require_relative "floe/workflow/runner/kubernetes"
-require_relative "floe/workflow/runner/podman"
 require_relative "floe/workflow/state"
 require_relative "floe/workflow/states/choice"
 require_relative "floe/workflow/states/fail"
@@ -33,6 +29,8 @@ require_relative "floe/workflow/states/pass"
 require_relative "floe/workflow/states/succeed"
 require_relative "floe/workflow/states/task"
 require_relative "floe/workflow/states/wait"
+
+require_relative "floe/runner/container_runner"
 
 require "jsonpath"
 require "time"
@@ -54,18 +52,5 @@ module Floe
   # @param logger [Logger] logger to use for logging actions
   def self.logger=(logger)
     @logger = logger
-  end
-
-  # Set the runner to use
-  #
-  # @example
-  #   Floe.set_runner "docker", kubernetes", {}
-  #   Floe.set_runner "docker", Floe::Workflow::Runner::Kubernetes.new({})
-  #
-  # @param scheme [String] scheme                           Protocol to register (e.g.: docker)
-  # @param name_or_instance [String|Floe::Workflow::Runner] Name of runner to use for docker (e.g.: docker)
-  # @param options [Hash]                                   Options for constructor of the runner (optional)
-  def self.set_runner(scheme, name_or_instance, options = {})
-    Floe::Workflow::Runner.set_runner(scheme, name_or_instance, options)
   end
 end
--- a/lib/floe/runner/container_runner.rb
+++ b/lib/floe/runner/container_runner.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+require_relative "container_runner/docker_mixin"
+require_relative "container_runner/docker"
+require_relative "container_runner/kubernetes"
+require_relative "container_runner/podman"
+
+module Floe
+  class Runner
+    class ContainerRunner
+      class << self
+        def cli_options(optimist)
+          optimist.banner("")
+          optimist.banner("Container runner options:")
+
+          optimist.opt :container_runner, "Type of runner for docker container images (docker, podman, or kubernetes)", :type => :string, :short => 'r'
+          optimist.opt :container_runner_options, "Options to pass to the container runner", :type => :strings, :short => 'o'
+
+          optimist.opt :docker,     "Use docker to run container images     (short for --container-runner=docker)",     :type => :boolean
+          optimist.opt :podman,     "Use podman to run container images     (short for --container-runner=podman)",     :type => :boolean
+          optimist.opt :kubernetes, "Use kubernetes to run container images (short for --container-runner=kubernetes)", :type => :boolean
+        end
+
+        def resolve_cli_options!(opts)
+          # shortcut support
+          opts[:container_runner] ||= "docker" if opts[:docker]
+          opts[:container_runner] ||= "podman" if opts[:podman]
+          opts[:container_runner] ||= "kubernetes" if opts[:kubernetes]
+
+          runner_options = opts[:container_runner_options].to_h { |opt| opt.split("=", 2) }
+
+          begin
+            set_runner(opts[:container_runner], runner_options)
+          rescue ArgumentError => e
+            Optimist.die(:container_runner, e.message)
+          end
+        end
+
+        def runner
+          @runner || set_runner(nil)
+        end
+
+        def set_runner(name_or_instance, options = {})
+          @runner =
+            case name_or_instance
+            when "docker", nil
+              Floe::Runner::ContainerRunner::Docker.new(options)
+            when "podman"
+              Floe::Runner::ContainerRunner::Podman.new(options)
+            when "kubernetes"
+              Floe::Runner::ContainerRunner::Kubernetes.new(options)
+            when Floe::Workflow::Runner
+              name_or_instance
+            else
+              raise ArgumentError, "container runner must be one of: docker, podman, kubernetes"
+            end
+        end
+      end
+    end
+  end
+end
+
+Floe::Workflow::Runner.register_scheme("docker", -> { Floe::Runner::ContainerRunner.runner })
--- a/lib/floe/runner/container_runner/docker.rb
+++ b/lib/floe/runner/container_runner/docker.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 module Floe
-  class Workflow
-    class Runner
+  class Runner
+    class ContainerRunner
       class Docker < Floe::Workflow::Runner
         include DockerMixin
 
--- a/lib/floe/workflow/runner.rb
+++ b/lib/floe/workflow/runner.rb
@@ -12,35 +12,21 @@ module Floe
 
       @runners = {}
       class << self
-        # deprecated -- use Floe.set_runner instead
-        def docker_runner=(value)
-          set_runner("docker", value)
+        def register_scheme(scheme, klass_or_proc)
+          @runners[scheme] = klass_or_proc
         end
 
-        # see Floe.set_runner
-        def set_runner(scheme, name_or_instance, options = {})
-          @runners[scheme] =
-            case name_or_instance
-            when "docker", nil
-              Floe::Workflow::Runner::Docker.new(options)
-            when "podman"
-              Floe::Workflow::Runner::Podman.new(options)
-            when "kubernetes"
-              Floe::Workflow::Runner::Kubernetes.new(options)
-            when Floe::Workflow::Runner
-              name_or_instance
-            else
-              raise ArgumentError, "docker runner must be one of: docker, podman, kubernetes"
-            end
+        private def resolve_scheme(scheme)
+          runner = @runners[scheme]
+          runner = @runners[scheme] = @runners[scheme].call if runner.is_a?(Proc)
+          runner
         end
 
         def for_resource(resource)
           raise ArgumentError, "resource cannot be nil" if resource.nil?
 
-          # if no runners are set, default docker:// to docker
-          set_runner("docker", "docker") if @runners.empty?
           scheme = resource.split("://").first
-          @runners[scheme] || raise(ArgumentError, "Invalid resource scheme [#{scheme}]")
+          resolve_scheme(scheme) || raise(ArgumentError, "Invalid resource scheme [#{scheme}]")
         end
       end
 
--- a/exe/floe
+++ b/exe/floe
@@ -13,33 +13,21 @@ opts = Optimist.options do
   opt :context, "JSON payload of the Context",                  :type => :string
   opt :credentials, "JSON payload with credentials",            :type => :string
   opt :credentials_file, "Path to a file with credentials",     :type => :string
-  opt :docker_runner, "Type of runner for docker images",       :type => :string, :short => 'r'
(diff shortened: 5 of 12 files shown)


lib/floe/runner/container_runner.rb
```diff
@@ -0,0 +1,63 @@
...
+
+          optimist.opt :container_runner, "Type of runner for docker container images (docker, podman, or kubernetes)", :type => :string, :short => 'r'
+          optimist.opt :container_runner_options, "Options to pass to the container runner", :type => :strings, :short => 'o'
+
+          optimist.opt :docker,     "Use docker to run container images     (short for --container_runner=docker)",     :type => :boolean
+          optimist.opt :podman,     "Use podman to run container images     (short for --container_runner=podman)",     :type => :boolean
+          optimist.opt :kubernetes, "Use kubernetes to run container images (short for --container_runner=kubernetes)", :type => :boolean
+        end
+
+        def resolve_cli_options(opts)
```
agrare, line 24: Looks like this modifies the opts being passed in, how about either `resolve_cli_options!` or maybe call this `set_container_runner`

Fryguy: oh yeah good idea! on the `!`

I wanted a generically named interface that could be called from exe/floe.  So, however the plugin wants to deal with the cli options is up to them.

Fryguy:
I know this changes the interface to exe/floe, and I'm not sure we wanted that.  I did it more to clarify that the runner is around running containers.  However, since the scheme is docker, we may want to continue to call it docker_runner for consistency.  That being said, I don't think we need to enforce that a plugin's name match the scheme, because that would preclude plugins that support multiple schemes.

lib/floe.rb
```diff
@@ -55,17 +53,4 @@ def self.logger
   def self.logger=(logger)
     @logger = logger
   end
-
-  # Set the runner to use
-  #
-  # @example
-  #   Floe.set_runner "docker", kubernetes", {}
-  #   Floe.set_runner "docker", Floe::Workflow::Runner::Kubernetes.new({})
```
kbrock, line 63: can we keep this as the public interface?

lib/floe/runner/container_runner.rb
```diff
@@ -0,0 +1,63 @@
...
+require_relative "container_runner/docker_mixin"
+require_relative "container_runner/docker"
+require_relative "container_runner/kubernetes"
+require_relative "container_runner/podman"
+
+module Floe
+  class Runner
+    class ContainerRunner
+      class << self
+        def cli_options(optimist)
```
kbrock, line 12: the original code had the user interface (cli) separate from the actual code
Does it make sense to keep these separate?

lib/floe/runner/container_runner/docker.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 module Floe
-  class Workflow
-    class Runner
+  class Runner
+    class ContainerRunner
```
kbrock, line 5: I thought the beauty of the old interface was that all runners, regardless of the scheme had the same runner interface.
The task calls into any runner, just picking the runner based upon the scheme.

So a task would work with a bash or jq runner or a "docker" runner (yes, this has multiple possible backends. And the same asl will work with any of the backends so you can use one backend for developemant and production with the same asl file)

So I was thinking that we wouldn't need to distinguish between a Container Runner and any other runner. Changing `Floe::Workflow` to `Floe::Runner` or `Floe::Runner::Base` makes sense to me (prefer the shorter form)

lib/floe/workflow/runner.rb
```diff
@@ -12,35 +12,21 @@ def initialize(_options = {})
...
-              Floe::Workflow::Runner::Kubernetes.new(options)
-            when Floe::Workflow::Runner
-              name_or_instance
-            else
-              raise ArgumentError, "docker runner must be one of: docker, podman, kubernetes"
-            end
+        private def resolve_scheme(scheme)
+          runner = @runners[scheme]
+          runner = @runners[scheme] = @runners[scheme].call if runner.is_a?(Proc)
+          runner
```
kbrock, line 22: The old `set_runner` wasn't robust enough. You will note that `providers-workflows` is basically this but with a bunch of special cases for the options hash.

Is it possible to move some of the provider special logic into here? (possibly getting rid of that?)

lib/floe/runner/container_runner.rb
kbrock: also, can you add a dummy bash/echo runner or something?
Think this will let us know if this `cli_options` introduction actually does simplify things.

lib/floe/runner/container_runner.rb
```diff
@@ -0,0 +1,63 @@
...
+            else
+              raise ArgumentError, "container runner must be one of: docker, podman, kubernetes"
+            end
+        end
+      end
+    end
+  end
+end
+
+Floe::Workflow::Runner.register_scheme("docker", -> { Floe::Runner::ContainerRunner.runner })
```
kbrock, line 63: I like the idea of also adding
```
Floe.register_plugin(Floe::Runner::ContainerRunner.new)
```

yes, the methods don't need to be an instance, but makes it easier to be called.
Hmm. this will work if we just pass in the class, since it will call the method on anything passed in.

Also like register_scheme to be called from `set_runner` - but that is a tomato/potatoe kind of thing. (agree to disagree and not my hill)

lib/floe.rb
```diff
@@ -34,6 +30,8 @@
 require_relative "floe/workflow/states/task"
 require_relative "floe/workflow/states/wait"
 
+require_relative "floe/runner/container_runner"
```
kbrock, line 33: cool thing here.

If you extract the gem.
and you put `gem "floe-container_runner", :require => true` into the `Gemfile`, then you can remove this line and ruby will register the plugin without much fuss.

kbrock:
> I actually wanted `Floe::ContainerRunner` directly, but didn't want to deal with the un-indenting of 1 level of namespace. I still may do this anyway, because then this would mirror what a real "plugin" would look like allowing for things like `require 'floe/container_runner'` and living in a floe-container_runner gem.

I like the unindent idea (and agree, not necessary to do it at this time).
That said, I still feel that the runner's parents should be `Floe::Runner` and not 

> While the exe/floe hardcodes the reference to the container runner, I think it's cleaner to loop over all of the registered schemes, and then call cli_options/resolve_cli_options on each. However, I thought it would make this PR less clear and didn't do that for now.

I mentioned above, I feel we need to register a plugin separate from registering the schemes.

lib/floe/runner/container_runner.rb
Fryguy: Maybe?  plugins should own their own code and that includes the cli options in my opinion.  Whether it's in floe/runner/container_runner.rb vs floe/runner/container_runner/cli.rb, I could go either way.  For this PR, I'd keep it as is, and that could be a future cleanup, IMO.

lib/floe.rb
Fryguy: `Floe.set_runner` essentially moved to `Floe::Workflow::Runner.register_scheme`.

There are 2 "APIs" here that I intentionally broke apart. There is an API for a plugin author and what they need to do to register a plugin.  Separately, there's an end-user API where end users can set things like Loggers.  IMO, the `Floe::Workflow::Runner` (or perhaps just `Floe::Runner`) namespace should be for the former, and the `Floe` namespace should be for the latter.

lib/floe/runner/container_runner/docker.rb
Fryguy: `ContainerRunner` naming here was intentional.  I was concerned naming it `DockerRunner` and that there might be confusion around the scheme vs the underlying runner choice.  That is one of the problems with our docker scheme is this particular confusion.  I am perfectly happy to change this back to DockerRunner to match the schema name.

lib/floe/workflow/runner.rb
Fryguy: > Is it possible to move some (all?) of the provider special logic into here?

Some, yes, however the _options_ are going to be significantly different so you'll still need that case statement in the workflows plugin.

lib/floe/runner/container_runner.rb
Fryguy: I considered something like that but it just felt like another level of unnecessary indirection, and this made it seems a lot more explicit what was happening.  This is something we can do in a followup because it's the plugin's API and there's only 1 plugin at the moment, so it's a an easy change if that's what we want.

lib/floe.rb
Fryguy: Yes exactly.  I actually considered dropping this line entirely and just having exe/floe do the require directly, but didn't want to rock the boat too much.

Fryguy:
> I mentioned above, I feel we need to register a plugin separate from registering the schemes.

Not sure I understand this distinction.  The only purpose of a plugin is to bring schema runners.

lib/floe/runner/container_runner.rb
```diff
@@ -0,0 +1,63 @@
...
+            Optimist.die(:container_runner, e.message)
+          end
+        end
+
+        def runner
+          @runner || set_runner(nil)
+        end
+
+        def set_runner(name_or_instance, options = {})
+          @runner =
```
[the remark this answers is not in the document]
Fryguy: In the manageiq-providers-workflows where we have it will change like follows

```diff
- Floe.set_runner("docker", "kubernetes", options)
+ Floe::Runner::ContainerRunner.set_runner("kubernetes", options)
```

that

Fryguy:
Added to the OP that I think this closes #45

kbrock:
LGTM
@Fryguy are you all set here?

Fryguy:
yup :+1:

lib/floe/runner/container_runner/docker.rb
agrare: > I thought the beauty of the old interface was that all runners, regardless of the scheme had the same runner interface.
> The task calls into any runner, just picking the runner based upon the scheme.

Yeah I agree, I was thinking of this like the active_record connection_adapters where there is an abstract_adapter that defines the interface and each plugin implements that interface.

Fryguy: I didn't really change that interface, so it still follows that pattern.  The ContainerRunner just happens to resolve the specific class dynamically based on configuration.  I guess it's very similar to the active record jdbc adapter, which talks jdbc, and part of the configuration is telling it which database to connect to through jdbc, whether it's mysql, postgresql, etc.  That latter connection type is resolved later.

agrare: >  The ContainerRunner just happens to resolve the specific class dynamically based on configuration

Right I'm not talking about that part of the dynamic resolver, I'm thinking about Task looking up the runner from the scheme and calling the Runner.  We can take docker/podman/kubernetes out of it for this and think about an Ansible runner that has to implement an interface in order to be used by Task.

lib/floe/runner/container_runner/docker.rb
```diff
@@ -1,8 +1,8 @@
...
       class Docker < Floe::Workflow::Runner
```
agrare, line 6: Oh I thought you were getting rid of Floe::Workflow::Runner based on the above comment, nevermind then

lib/floe/runner/container_runner/docker.rb
agrare: Ignore above, I thought the runner wasn't inheriting from the base runner class anymore

agrare (approved):

Fryguy:
After this PR is merged I'll do some of the indent changes as discussed.

lib/floe/runner/container_runner.rb
```diff
@@ -0,0 +1,63 @@
...
+        end
+
+        def resolve_cli_options!(opts)
+          # shortcut support
+          opts[:container_runner] ||= "docker" if opts[:docker]
+          opts[:container_runner] ||= "podman" if opts[:podman]
+          opts[:container_runner] ||= "kubernetes" if opts[:kubernetes]
+
+          runner_options = opts[:container_runner_options].to_h { |opt| opt.split("=", 2) }
+
```
kbrock, line 31: accepting options and how the options are formatted is an optimist problem.

So I think that is part of the cli layer

Handling adding a container runner name parameter is very container runner. so it belongs as part of the container runner optimist extension. (how this extension enhances the ui)

Not sure the final answer. So that means this is good enough

Fryguy: Yeah I agree - the "common" cli options parsing can be pushed down to a shared method.  I can most definitely do that in a follow up.

PR #174, opened 2024-04-16
Fix issues with exe/floe and various combinations of workflow and input

Fryguy:
This commit cleans up the exe/floe interface for bare workflow/input pairs when passed in combination with `--workflow` and `--input`. This commit also makes the `--workflow` and `--input` params no longer legacy but supported.

The following are now all accepted ways to pass these parameters:

```sh
# Single workflow
exe/floe workflow1.asl
exe/floe workflow1.asl '{"input":1}'
exe/floe workflow1.asl --input '{"input":1}'
exe/floe --workflow workflow1.asl --input '{"input":1}'

# Multiple workflows
exe/floe workflow1.asl '{"input":1}' workflow2.asl '{"input":2}'
exe/floe workflow1.asl workflow2.asl --input '{"input":1}' # input is dup'd for each workflow
```

The following will raise errors:

```sh
exe/floe workflow1.asl workflow2.asl               # ambiguous as to workflow/input pair or 2 workflows, and so requires input
exe/floe workflow1.asl --workflow workflow2.asl    # invalid mixing of bare workflows and parameter
exe/floe workflow1.asl workflow2.asl '{"input":1}' # mismatched workflow/input pairs
exe/floe --input '{"input":1}'                     # missing workflow
exe/floe                                           # no parameters
```

### DIFF
--- a/spec/exe/floe_spec.rb
+++ b/spec/exe/floe_spec.rb
@@ -1,14 +1,122 @@
-RSpec.describe "exe/floe" do
+RSpec.describe "exe/floe", :slow => true do
   include AwesomeSpawn::SpecHelper
   before { enable_spawning }
 
-  it "displays help" do
-    output = AwesomeSpawn.run!("exe/floe --help").output
+  let(:exe)      { File.expand_path("../../exe/floe", __dir__) }
+  let(:workflow) { File.expand_path("data/workflow.asl", __dir__) }
 
-    expect(output).to start_with("Usage: floe [options] workflow input [workflow2 input2]")
-    expect(output).to include(Floe::VERSION)
+  it "displays help" do
+    output = AwesomeSpawn.run!(exe, :params => [:help]).output
+    lines = output.lines(:chomp => true)
+
+    expect(lines).to start_with("Usage: floe [options] workflow input [workflow2 input2]")
+    expect(lines).to include("v#{Floe::VERSION}")
 
     # it should also include options from runners
-    expect(output).to include("Container runner options:")
+    expect(lines).to include("Container runner options:")
+  end
+
+  it "with no parameters" do
+    result = AwesomeSpawn.run(exe)
+    expect(result).to be_failure
+
+    lines = result.error.lines(:chomp => true)
+    expect(lines.first).to eq("Error: workflow/input pairs must be specified.")
+  end
+
+  it "with a bare workflow and no input" do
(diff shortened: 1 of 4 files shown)


spec/exe/floe_spec.rb
```diff
@@ -1,14 +1,122 @@
-RSpec.describe "exe/floe" do
+RSpec.describe "exe/floe", :slow => true do
```
Fryguy, line 1: Thought this might be useful, since these tests are kinda slow, so they can be ignored in local dev with `--tag ~slow`

spec/exe/floe_spec.rb
```diff
@@ -1,14 +1,122 @@
...
+
+    expect(lines.first).to include("checking 2 workflows...")
+    expect(lines.last(7).join("\n")).to eq(<<~OUTPUT.chomp)
+      workflow
+      ===
+      {"foo"=>1}
+
+      workflow
+      ===
+      {"foo"=>2}
```
Fryguy, line 95: This was surprising to me (here and of course everywhere else), as I assumed the output would be JSON formatted and not Ruby formatted.  So, the question is, should it be JSON?  If so, it's an easy fix for a followup PR.

Fryguy: I also expected either the workflow.asl file name or perhaps the workflow comment (or both), so just the word "workflow" was also surprising.

kbrock: I modified this to run multiple workflows.
Goals:

- run multiple providers.
- run in a way similar to how the provider runs
- allow easy migration to the non-polling code.
- easily determine that all workflows were run and their output status.

Modifying it sounds great.

kbrock:
also, can we drop `--input` and `--workflow`.
I left them in when introducing ARGV, but think it is time to drop them.

Thoughts?

Fryguy:
No this PR makes them a) permanent (I use them personally and prefer them) and b) input, specifically, is useful for multiple workflows that take the same input cause you only need to specify it once

spec/exe/floe_spec.rb
Fryguy: I wasn't really thinking about post processing, but that would be useful in the single workflow case.  Even so, it was more of an expectation thing...for some reason I expected a json output.

Fryguy:
I could see maybe removing --workflow in a future PR, but --input has a use case.

agrare (approved):

PR #176, opened 2024-04-30
drop run!

kbrock:
## Overview

- We dropped all references to `run!` in 46c175f ( https://github.com/ManageIQ/floe/pull/149 ) but did not drop it from the abstract interface: `Floe::Runner#run!`
- `Task#running?` called `status!` but it was not part of the abstract interface: `Floe::Runner#status!`.
- 2 documentation changes -- Wondering if we should remove the leading `_` from method parameters and add yarddoc for all parameters.

## After

- Drop `Floe::Runner#run!`
- Add `Floe::Runner#status!`
- Added yarddoc for `Floe::Runner` methods
- renamed parameter `run_async!(image, ...)` to `run_async!(resource, ...)`

## Alternatives

If we don't want `Floe::Runner#status!` as part of the interface, then we can move that call from `Task#running?` to the `Floe::ContainerRunner::*#running?` implementations.

### DIFF
--- a/lib/floe/runner.rb
+++ b/lib/floe/runner.rb
@@ -29,10 +29,6 @@ module Floe
       end
     end
 
-    def run!(resource, env = {}, secrets = {})
-      raise NotImplementedError, "Must be implemented in a subclass"
-    end
-
     # @return [Hash] runner_context
     def run_async!(_image, _env = {}, _secrets = {})
       raise NotImplementedError, "Must be implemented in a subclass"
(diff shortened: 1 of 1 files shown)


kbrock:
update:
- added method `status!` to the interface.
- added yarddoc

kbrock:
@agrare @Fryguy Curious if you both want `status!` as part of the interface or want to make `running?` a method that can change the `runner_context`.

Almost feel like it could be renamed to `check_status!`
And calling it from `Task#running?`

kbrock:
update:
- renamed `run_async!(image, ...)` to `run_async!(resource, ...) - no real change

agrare (approved):

PR #177, opened 2024-04-30
pass task context into container_runners

kbrock:
This allows runners to access the execution context of a request

want input:
Debating passing the whole `context` vs just `context.execution`.


lib/floe/workflow/states/task.rb
```diff
@@ -36,7 +36,7 @@ def start(input)
           super
 
           input          = process_input(input)
-          runner_context = runner.run_async!(resource, input, credentials&.value({}, workflow.credentials))
+          runner_context = runner.run_async!(resource, input, credentials&.value({}, workflow.credentials), context.execution)
```
agrare, line 39: Could we pass the whole context in and then let the runner decide what out of that to pass along?  Thinking that context.current_state could be helpful as well.

Not necessarily here but we should probably enforce that this isn't to be modified by callers (e.g. a built-in method shouldn't be able to set anything in context.execution), and can only set output

kbrock:
update:
- passing full context into the runner

agrare (approved):

PR #178, opened 2024-04-30
[WIP] [POC] Awesome runner

kbrock:
depends upon:
- [x] #177 **First Commit**

## Overview

- Current runners link to an external process that manages the runners.
- Current runners are asynchronous

So that begs to ask the question on how to implement a runner that manages its own subprocesses.

## Implementation

I came up with a synchronous and an asynchronous implementation. They get converted across with a commit.

I feel the synchronous implementation may be the best one, otherwise it has a bunch of threads hanging around.

commits:

- dependency
- Synchronous implementation
- asynchronous implementation

## Issues

This doesn't work perfectly since `AwesomeSpawn` escapes `$VAR` and that makes this pretty useless.

But the goal was more focused on the Threads and stuff rather than the actual awesome spawn part.

### DIFF
--- a/lib/floe/awesome_runner.rb
+++ b/lib/floe/awesome_runner.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+module Floe
+  class AwesomeRunner < Floe::Runner
+    SCHEME        = "awesome"
+    SCHEME_PREFIX = "#{SCHEME}://"
+    SCHEME_OFFSET = SCHEME.length + 3
+
+    def initialize(_options = {})
+      require "awesome_spawn"
+
+      super
+    end
+
+    # @return [Hash] runner_context
+    def run_async!(resource, params = {}, _secrets = {}, _context = {})
+      raise ArgumentError, "Invalid resource" unless resource&.start_with?(SCHEME_PREFIX)
+
+      args = resource[SCHEME_OFFSET..].split
+      method = args.shift
+
+      runner_context = {}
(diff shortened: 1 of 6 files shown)


kbrock:
updates:
- fixed wait code to run for multiple schemes
- many cops

kbrock:
update:
- fix comments around multi threading

lib/floe/awesome_runner.rb
```diff
@@ -0,0 +1,130 @@
+# frozen_string_literal: true
+
+require "concurrent/array"
+
+module Floe
+  class AwesomeProcess < Thread
```
kbrock, line 6: would have preferred not using a thread per process and not using an `AwesomeSpan` centric model.

kbrock:
update:
- rebase

update:
- fixed spec (due to rebase)

kbrock:
update:
- rebased

still WIP

Fryguy:
naming-wise I don't like the awesome part - externally awesome:// means nothing.  If this is basically just shell execution, let's call it shell:// or if it's bash specific then bash://

Fryguy:
I'm considering bringing this in as an optional built-in runner, but I want to refactor the way we do built-in runners first.  I have some ideas I want to write up.

kbrock:
I was really frustrated that variables could not be passed into the shell to pick up on the env. But I guess using ruby variables instead may fix for many of the use cases

PR #179, opened 2024-05-01
Implement runner#wait in a generic way

kbrock:
pulled out of #178

Depend upon:
- [x] #190

## Overview

The `builtin` runner does not need a `wait` method since it does not listen to completed tasks. But other runners may need a a `wait`.

So the solution is to make wait optional, and setup a thread for all the runners that implement that interface.

## Before

Run a thread to support `wait` from `"docker"` runner, but no others.

## After

Run a thread for every runner that requires `wait`.

Developer speak
===============

```ruby
# handle docker scheme
Thread.new do
  loop do
    Runner.for_resource("docker").wait do |event, runner_context|
      queue.push([event, runner_context])
    end
  end
end
```

aka:

```ruby
# handle docker scheme
runner = Runner.for_resource("docker")
Thread.new do
  loop do
    runner.wait do |event, runner_context|
      queue.push([event, runner_context])
    end
  end
end
```

handling multiple runners:

```ruby
# handle all schemes
Runner.runners.each do |runner|
  Thread.new do
    loop do
      runner.wait do |event, runner_context|
        queue.push([event, runner_context])
      end
    end
  end
end
```

## Aside

I had wanted to add something with `exe/floe`'s interface, since that is also hard coded, but this addresses my primary concerns.

### DIFF
--- a/lib/floe/runner.rb
+++ b/lib/floe/runner.rb
@@ -27,6 +27,13 @@ module Floe
         scheme = resource.split("://").first
         resolve_scheme(scheme) || raise(ArgumentError, "Invalid resource scheme [#{scheme}]")
       end
+
+      def runners
+        @runners.each_value.map do |runner|
+          runner = runner.call if runner.kind_of?(Proc)
+          runner
+        end
+      end
     end
 
     # Run a command asynchronously and create a runner_context
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -20,16 +20,21 @@ module Floe
         workflows = [workflows] if workflows.kind_of?(self)
         logger.info("checking #{workflows.count} workflows...")
 
-        run_until   = Time.now.utc + timeout if timeout.to_i > 0
-        ready       = []
-        queue       = Queue.new
-        wait_thread = Thread.new do
-          loop do
-            Runner.for_resource("docker").wait do |event, runner_context|
-              queue.push([event, runner_context])
+        run_until    = Time.now.utc + timeout if timeout.to_i > 0
+        ready        = []
+        queue        = Queue.new
+        wait_threads =
+          Runner.runners.map do |runner|
+            next unless runner.respond_to?(:wait)
+
+            Thread.new do
+              loop do
+                runner.wait do |event, runner_context|
+                  queue.push([event, runner_context])
+                end
+              end
             end
           end
-        end
 
         loop do
           ready = workflows.select(&:step_nonblock_ready?)
@@ -81,7 +86,7 @@ module Floe
         logger.info("checking #{workflows.count} workflows...Complete - #{ready.count} ready")
         ready
       ensure
-        wait_thread&.kill
+        wait_threads.compact.map(&:kill)
       end
     end
 


lib/floe/workflow.rb
```diff
@@ -20,16 +20,21 @@ def wait(workflows, timeout: nil, &block)
...
-        wait_thread = Thread.new do
-          loop do
-            Runner.for_resource("docker").wait do |event, runner_context|
-              queue.push([event, runner_context])
+        run_until    = Time.now.utc + timeout if timeout.to_i > 0
+        ready        = []
+        queue        = Queue.new
+        wait_threads =
+          Runner.runners.map do |runner|
+            next unless runner.respond_to?(:wait)
```
agrare, line 28: @kbrock if we don't have a wait thread what is going to unblock the `queue.pop`
```
            # Block until an event is raised
            event, runner_context = queue.pop
```

kbrock: @agrare That thread is only needed to handle the async nature of the `"docker"` runner. I commented out the `"docker"` thread and was able to run a workflow just fine:

```json
{
  "Comment": "Directory Listing",
  "StartAt": "a",
  "States": {
    "a":{
      "Type": "Pass",
      "Next": "b"
    },
    "b": {
      "Type": "Wait",
      "Seconds": 1,
      "Next": "c"
    },
      "c": {
        "Type": "Succeed"
    }
  }
}
```

ASIDE: was also able to call with an `awesome://` `Task` without this thread

lib/floe/runner.rb
```diff
@@ -17,7 +17,7 @@ def register_scheme(scheme, klass_or_proc)
 
       private def resolve_scheme(scheme)
         runner = @runners[scheme]
-        runner = @runners[scheme] = @runners[scheme].call if runner.is_a?(Proc)
+        runner = @runners[scheme] = @runners[scheme].call if runner.kind_of?(Proc)
```
kbrock, line 20: this is a lazy eval, but only the first time hit.
I do wonder if this is not thread safe. Ruby's GPL probably takes care of it.

lib/floe/runner.rb
```diff
@@ -27,6 +27,13 @@ def for_resource(resource)
         scheme = resource.split("://").first
         resolve_scheme(scheme) || raise(ArgumentError, "Invalid resource scheme [#{scheme}]")
       end
+
+      def runners
+        @runners.each_value.map do |runner|
+          runner = runner.call if runner.kind_of?(Proc)
+          runner
+        end
+      end
```
kbrock, line 36: This is essentially `@runners.values` resolving the `procs`.

---

I feel supporting multiple schemes by a single runner is currently far fetched.
Ignoring that use case allows us to keep to our current simple runner registration process: `Floe::Runner.register_scheme`.

In the future, if we come up with a case where there are multiple schemes, we can make our registration process more complicated.

kbrock:
@Fryguy re: use case

I had implemented https://github.com/ManageIQ/floe/pull/178 using a synchronous/blocking implementation (first commit) and an async/event implementation (second commit)

The sync/blocking implementation works fine without this PR.
But if we have events triggered once the process is complete, then we need the `wait` thread to process these events.

This PR provides the thread for any runner that needs one.

lib/floe/runner.rb
```diff
@@ -75,8 +82,9 @@ def cleanup(_runner_context)
       raise NotImplementedError, "Must be implemented in a subclass"
     end
 
-    def wait(timeout: nil, events: %i[create update delete])
-      raise NotImplementedError, "Must be implemented in a subclass"
-    end
+    # Implement if the runner needs a watch for async events
+    # def wait(timeout: nil, events: %i[create update delete])
+    #   raise NotImplementedError, "Must be implemented in a subclass"
+    # end
```
kbrock, line 88: some runners do not have async events.
Putting an empty method into wait calls an empty method in a tight loop.
So having a switch that says a thread is not needed seems the best route (i.e.: not defining `wait`.

kbrock:
reducing branches: handle this with awesome_spawn

PR #182, opened 2024-05-13
A failed state can leave a workflow in "running"

agrare:
Right now just tweaking the specs to show the issue.

The problem is `Task#finish` has the following in the event of a state failure
https://github.com/ManageIQ/floe/blob/master/lib/floe/workflow/states/task.rb#L51-L56
```ruby
          else
            context.next_state = nil
            context.output = error = parse_error(output)
            super
            retry_state!(error) || catch_error!(error) || fail_workflow!(error)
          end
```

But `super` goes to the `NonTerminalMixin` https://github.com/ManageIQ/floe/blob/master/lib/floe/workflow/states/non_terminal_mixin.rb#L7-L10 which does:
```ruby
        def finish
          context.next_state = end? ? nil : @next
          super
        end
```

If the state doesn't have `"End": true` in the payload, this will re-set the next_state to `"Next": ...` and not mark the workflow as failed&ended here: https://github.com/ManageIQ/floe/blob/master/lib/floe/workflow/state.rb#L71
```ruby
        context.execution["EndTime"]    = finished_time_iso if context.next_state.nil?
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Fixed
+- Fix issue where a failed state can leave a workflow in "running" ([#182](https://github.com/ManageIQ/floe/pull/182))
 
 ## [0.11.0] - 2024-05-02
 ### Fixed
--- a/lib/floe/workflow/states/non_terminal_mixin.rb
+++ b/lib/floe/workflow/states/non_terminal_mixin.rb
@@ -4,11 +4,6 @@ module Floe
   class Workflow
     module States
       module NonTerminalMixin
-        def finish
-          context.next_state = end? ? nil : @next
-          super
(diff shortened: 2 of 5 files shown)


lib/floe/workflow/states/pass.rb
```diff
@@ -28,6 +28,7 @@ def finish
           input              = process_input(context.input)
           context.output     = process_output(input, result)
           context.next_state = end? ? nil : @next
+          context.next_state = end? ? nil : @next
```
kbrock, line 31: oops

agrare: Thanks, was just trying to get something working last night :sweat_smile:

kbrock:
darn. really liked setting these values in finished.
At finished, we are done with them.

ok. but the output and next state seems to still look good in the logs.

agrare:
> darn. really liked setting these values in finished.

Well it still is set in finished, just not in the mixin.

kbrock:
this is great

PR #186, opened 2024-05-22
Extract ErrorMatcherMixin from Catch and Retry

kbrock:
- Extract ErrorMatcherMixin#match_error?
- `States.HeartbeatTimeout` is handled by `States.Timeout`
- `States.Runtime` is not caught by any catchers/retriers

### DIFF
--- a/lib/floe.rb
+++ b/lib/floe.rb
@@ -8,6 +8,7 @@ require_relative "floe/logging"
 require_relative "floe/runner"
 
 require_relative "floe/workflow"
+require_relative "floe/workflow/error_matcher_mixin"
 require_relative "floe/workflow/catcher"
 require_relative "floe/workflow/choice_rule"
 require_relative "floe/workflow/choice_rule/not"
--- a/lib/floe/workflow/catcher.rb
+++ b/lib/floe/workflow/catcher.rb
@@ -3,6 +3,8 @@
 module Floe
   class Workflow
     class Catcher
+      include Floe::Workflow::ErrorMatcherMixin
+
       attr_reader :error_equals, :next, :result_path
 
       def initialize(payload)
@@ -11,6 +13,7 @@ module Floe
         @error_equals = payload["ErrorEquals"]
         @next         = payload["Next"]
         @result_path  = ReferencePath.new(payload.fetch("ResultPath", "$"))
+        raise Floe::InvalidWorkflowError, "State requires ErrorEquals" if !@error_equals.kind_of?(Array) || @error_equals.empty?
       end
     end
   end
--- a/lib/floe/workflow/error_matcher_mixin.rb
+++ b/lib/floe/workflow/error_matcher_mixin.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Floe
+  class Workflow
+    # Methods for common error handling
+    module ErrorMatcherMixin
+      # @param [String] error the error thrown
+      def match_error?(error)
+        return true if error_equals.include?("States.ALL")
+
+        error_equals.include?(error)
+      end
+    end
+  end
+end
--- a/lib/floe/workflow/error_matcher_mixin.rb
+++ b/lib/floe/workflow/error_matcher_mixin.rb
@@ -7,6 +7,7 @@ module Floe
       # @param [String] error the error thrown
       def match_error?(error)
         return true if error_equals.include?("States.ALL")
+        return true if error_equals.include?("States.Timeout") && error == "States.HeartbeatTimeout"
 
         error_equals.include?(error)
       end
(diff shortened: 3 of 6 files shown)


Fryguy:
I think your branch name and PR title are wrong?

lib/floe/workflow/error_matcher.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Floe
+  class Workflow
+    module ErrorMatcher
+      # note, expecting error_equals from payload to be available
```
Fryguy, line 6: Since this comment is a requirement, I wouldn't call it a note - I think this can more clearly be written as a comment on the module:


```suggestion
module Floe
  class Workflow
    # Contains methods for common error handling
    #
    # Requires that the including class to have the `error_equals` from the payload to be available.
    module ErrorMatcher
      # note, expecting error_equals from payload to be available
```

or something like that.

Fryguy: Or perhaps it's just an addition to the comment for the `match_error?` method specifically.

Fryguy: Or perhaps it's a note, but using `@note` yard tag in the `match_error?` method's comment specifically.  I can't really find an `@requires` or `@depends` yard tag.

kbrock:
update:
- fixed note
- fixed specs (was not properly comparing `equals_value`)

kbrock:
un-wip: tests passing

lib/floe/workflow/error_matcher.rb
```diff
@@ -0,0 +1,19 @@
...
+module Floe
+  class Workflow
+    # Methods for common error handling
+    #
+    # @note Included class required to have `error_equals` available from the payload.
+    module ErrorMatcher
+      # @param [String] error the error thrown
+      def match_error?(error)
+        return false if error == "States.Runtime"
+        return true if error_equals.include?("States.ALL")
```
Fryguy, line 12: Do we need to check that the author set error_equals correctly?  If it was unset, then I think this is nil, and if not set to an Array, this is a string include check?  (I think the old code also failed, but it was more subtle).

Should we add validation of the error equals payload?

kbrock: old code was:

```ruby
(r.error_equals & [error, "States.ALL"]).any?

nil & [1,2]
# => nil
```

So yea, it silently failed.

kbrock:
update:
- added error_equals validation

kbrock:
update:
- moved error_equals up to parent - no longer the `Note` stuff
- added "States.ALL" special case
- generalized Heartbeat special case
- specs for `ErrorMatcher`

lib/floe/workflow/error_matcher.rb
```diff
@@ -0,0 +1,34 @@
...
+    # @note Included class required to have `error_equals` available from the payload.
+    module ErrorMatcher
+      attr_reader :error_equals, :error_equals_all
+
+      def initialize(_workflow, name, payload)
+        @error_equals = payload["ErrorEquals"]
+        validate_error_equals!(name, ".#{payload["Type"]}")
+
+        @error_equals_all = error_equals.include?("States.ALL")
+        error_equals << "States.HeartbeatTimeout" if error_equals.include?("States.Timeout")
```
kbrock, line 16: I kept this optimization in a separate commit.

2 camps:
- modify this array and no special `if` cases.
- don't modify this array and add special `if` logic.

lib/floe/workflow/error_matcher.rb
kbrock: I was able to extract this into the parent so this is no longer needed.

lib/floe/workflow/error_matcher.rb
kbrock: Added validation so we catch a retry with an empty, missing, or non-array ErrorEquals

kbrock:
update:

- rewrote tests to go through workflow
- added tests for validation
- now checks that retrier (and catch) are non-empty

update:

- fixed `tap` in tests

kbrock:
update:
- fix code complexity via validate_array_field!

kbrock:
update:

- compressed _catch_all case to convert guard into just a `boolean ||`

kbrock:
update:
- rebased
- depend upon common error handling

kbrock:
update:
- rebase (fixed conflict)
- rubocop: fixed spacing

kbrock:
update:
- based upon #209 
- Greatly simplified the logic in here.

kbrock:
update:
- remove 209 - trimmed it down

kbrock:
update:
- rubocop

lib/floe/workflow/error_matcher.rb
kbrock: that was too much of a leap.
I went back to keeping this PR as just extraction and moving that into #212

kbrock:
update:
- extracted #212 (few pushes back) to make this as simple as I can

update:
- rebase master

kbrock:
@agrare @Fryguy this has been greatly simplified. Is this good to go now?

agrare (approved):

PR #191, opened 2024-05-23
output now based upon raw input not effective input

kbrock:
## Context

```
// raw input
{
  "title": "Numbers to add",
  "numbers": { "val1": 3, "val2": 4 }
}

// result
7

// state
"InputPath": "$.numbers",
"ResultPath": "$.sum"
```

## Before

```
// raw output:
{
  "val1": 3,
  "val2": 4,
  "sum": 7
}
```

## After

```
// raw output:
{
  "title": "Numbers to add",
  "numbers": { "val1": 3, "val2": 4 },
  "sum": 7
}
```

refs:
- https://states-language.net/#inputoutput-processing-examples
- https://docs.aws.amazon.com/step-functions/latest/dg/input-output-example.html

### DIFF
--- a/spec/workflow/states/pass_spec.rb
+++ b/spec/workflow/states/pass_spec.rb
@@ -53,5 +53,41 @@ RSpec.describe Floe::Workflow::States::Pass do
         expect(ctx.next_state).to eq("SuccessState")
       end
     end
+
+    # https://states-language.net/#inputoutput-processing-examples
+    context "with 2 blocks" do
+      let(:payload) do
+        {
+          "First"   => {
+            "Type"       => "Pass",
+            "Result"     => {
+              "title"   => "Numbers to add",
+              "numbers" => {"val1" => 3, "val2" => 4}
+            },
+            "ResultPath" => "$",
+            "Next"       => "Second"
+          },
+          "Second"  => {
+            "Type"       => "Pass",
+            "Result"     => 7,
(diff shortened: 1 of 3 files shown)


spec/workflow/states/pass_spec.rb
```diff
@@ -53,5 +53,41 @@
...
+            "ResultPath" => "$.sum",
+            "Next"       => "Success"
+          },
+          "Success" => {
+            "Type" => "Succeed"
+          }
+        }
+      end
+
+      it "Uses raw input not" do
```
Fryguy, line 83: Your text got cut off here.

Fryguy:
Will this break any of our existing workflows like the ones in workflows-examples, or the new ootb ones?

kbrock:
update:
- fixed spec title
- cops: fixed indentation

kbrock:
Searched through https://github.com/search?q=org%3AManageIQ%20path%3A*.asl&type=code (8 files)

and for agrare, fryguy

We don't use `InputPath` at all. so this should no affect us.

/cc @agrare - do you see any change that may be introduced if we are not using `InputPath` here?

That is probably why the bug has not been discovered yet

agrare (approved):
Great catch, LGTM

agrare:
@kbrock yeah agreed I don't see any use of InputPath in our existing examples

PR #192, opened 2024-05-23
require mfa

kbrock:
rubocop.

But I think we want to set this anyway


kbrock:
update:
- use a string for mfa

Fryguy (approved):

PR #193, opened 2024-05-23
reduce complexity ContainerRunner#resolve_cli_options

kbrock:
not sure this is necessary

But it the comment 'shortcut support' suggests putting this in a method works

### DIFF
--- a/lib/floe/container_runner.rb
+++ b/lib/floe/container_runner.rb
@@ -21,11 +21,17 @@ module Floe
       end
 
       def resolve_cli_options!(opts)
-        # shortcut support
+        resolve_shortcuts(opts)
+        resolve_options(opts)
+      end
+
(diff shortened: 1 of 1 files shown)


kbrock:
can't believe even introducing this simple method it is too "complex"

going forward ignoring the code climate issue around this method

Fryguy:
@kbrock This is what I was referring too where we bumped the value to higher than 5: https://github.com/ManageIQ/manageiq-style/blob/master/styles/cc_base.yml#L36-L38 .  I'm curious why this is not being honored in codeclimate.

kbrock:
@Fryguy it feels like the bash prompt `rubocop` follows slightly different rules than miq-bot

Fryguy:
That's correct, we have different ones for code climate (which only looks at metrics) and the bot (which only looks at style)

PR #194, opened 2024-05-24
InstrinsicFunction foundation + States.Array, States.UUID

Fryguy:
This is the start of a PR for implementing [Intrinsic Functions](https://states-language.net/#appendix-b) (re: #64) parsing with a parser generator.

Intrinsic functions are effectively raw Strings, and thus are their own "language" within that string.  As such, we need a proper parser. This PR uses the `parslet` gem , which is a super simple pure-Ruby parser generator.  Some other choices we can use are `rexical`+`racc`, `lrama`, `ragel`, `TreeTop`.  However, I chose `parslet` because it was super-simple to use, very fast, no dependencies, and pure Ruby.  I figure we can always change later once the parse rules are written, because conceptually they shouldn't change between gems (aside from syntax, of course), and the test suite is really the important part and should stay the same through any rewrites.

So far I've only implemented `States.UUID` for something super-simple, and `States.Array` for something relatively complex.  I will add the rest in a follow-up.

---

A quick primer on `parslet`.  ([docs](https://kschiess.github.io/parslet/))

Parselet breaks the parser routine into 2 parts.  The first part is the `Parslet::Parser`, where you define rules in Parsing Expression Grammar style.  The output of the parser is a tree.  However, what's cool with parslet is that within the parser you annotate what you want to appear in the tree, so you can pull out the important semantic stuff, and ignore the syntax stuff.  This is done using the `.as(symbol)` function, where the symbol is what will be the key in the tree.  Here is an example tree output from the parser:

```ruby
payload = "States.Array('string', 1, 1.5, true, false, null)"
tree = Floe::Workflow::IntrinsicFunction::Parser.new.parse(payload)
pp tree
# {:states_array=>
#   [{:string=>"string"@14},
#    {:number=>"1"@23},
#    {:number=>"1.5"@26},
#    {:true_literal=>"true"@31},
#    {:false_literal=>"false"@37},
#    {:null_literal=>"null"@44},
#    {:jsonpath=>"$.input"@50}]}
```

The second part is the `Parslet::Transform`, where you define rules on how to convert that tree into something usable.  It does so in a depth-first pattern-matching style, so you define transform rules based on the key names above, and then tell it how to process the values.  Here is an example of feeding the tree into the transformer:

```ruby
Floe::Workflow::IntrinsicFunction::Transformer.new.apply(tree, :input => {"input" => {"foo" => "bar"}})
# => ["string", 1, 1.5, true, false, nil, {"foo"=>"bar"}]
```

In this PR, I just put those together into a simple `Floe::Workflow::IntrinsicFunction.evaluate` method.


floe.gemspec
```diff
@@ -36,6 +36,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "jsonpath", "~>1.1"
   spec.add_dependency "kubeclient", "~>4.7"
   spec.add_dependency "optimist", "~>3.0"
+  spec.add_dependency "parslet"
```
agrare, line 39: Maybe throw a `"~>2.0" so we don't get any breaking surprises?

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -0,0 +1,31 @@
...
+
+        rule(:string   => simple(:v)) { v.to_s }
+        rule(:number   => simple(:v)) { v.match(/[eE.]/) ? Float(v) : Integer(v) }
+        rule(:jsonpath => simple(:v)) do
+          results = JsonPath.on(input, v.to_s)
+          results.count < 2 ? results.first : results
+        end
+
+        rule(:states_array => subtree(:args)) { args.kind_of?(Array) ? args : [args] }
+        rule(:states_uuid  => subtree(:args)) { SecureRandom.uuid }
```
agrare, line 25: Nice that looks really easy to add new functions

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -0,0 +1,31 @@
...
+      class Transformer < Parslet::Transform
+        rule(:null_literal  => simple(:v)) { nil }
+        rule(:true_literal  => simple(:v)) { true }
+        rule(:false_literal => simple(:v)) { false }
+
+        rule(:string   => simple(:v)) { v.to_s }
+        rule(:number   => simple(:v)) { v.match(/[eE.]/) ? Float(v) : Integer(v) }
+        rule(:jsonpath => simple(:v)) do
+          results = JsonPath.on(input, v.to_s)
+          results.count < 2 ? results.first : results
```
Fryguy, line 21: I think I want to change this to reuse Floe::Workflow::Path, but I might wait for a follow up PR.  This also doesn't handle the context at the moment.

spec/workflow/intrinsic_function/parser_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+# rubocop:disable Style/PercentLiteralDelimiters, Style/RedundantPercentQ
+
+require 'parslet/rig/rspec'
+
+RSpec.describe Floe::Workflow::IntrinsicFunction::Parser do
+  describe "spaces" do
+    subject { described_class.new.spaces }
+
+    it do
+      expect(subject).to parse(" ")
```
agrare, line 13: Is `expect().to parse()` just checking that it doesn't raise an exception?  Is there a parse result we could be checking?

spec/workflow/intrinsic_function/parser_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+      expect(subject).to parse(%q|$['store']['book']|)
+      expect(subject).to parse(%q|$['store'][0]['book']|)
+
+      # TODO: These unlikely cases should work, but we need a more thorough parser.
+      expect(subject).to_not parse(%q|$['s,tore']|)
+      expect(subject).to_not parse(%q|$['s)tore']|)
+
+      expect(subject).to_not parse(%q|['input']|)
+      expect(subject).to_not parse(".input")
+      expect(subject).to_not parse("$")
```
agrare, line 163: Hm this one is interesting, isn't `$` valid jsonpath for "return the entire input"

lib/floe/workflow/intrinsic_function.rb
```diff
@@ -0,0 +1,11 @@
+module Floe
+  class Workflow
+    module IntrinsicFunction
+      def self.evaluate(payload, input = {})
+        tree = Parser.new.parse(payload)
+        pp tree if ENV["DEBUG"]
```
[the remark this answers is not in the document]
agrare: If we keep it in would prefer to use the Floe logger

spec/workflow/intrinsic_function/parser_spec.rb
Fryguy: This tests the parser directly and not the transformer, so yeah, more or less.  The results from the parser are just the tree, which could be a useful set of tests especially when parsing something more complicated like nested functions, but here I just wanted validation of good/bad cases.

Here's some example results:

```ruby
$ bin/console
irb(main):001:0> Floe::Workflow::IntrinsicFunction::Parser.new.spaces.parse(" ")
=> " "@0
irb(main):002:0> Floe::Workflow::IntrinsicFunction::Parser.new.number.parse("123.456")
=> {:number=>"123.456"@0}
irb(main):003:0> Floe::Workflow::IntrinsicFunction::Parser.new.number.parse("xxx")
/Users/jfrey/.gem/ruby/3.1.5/gems/parslet-2.0.0/lib/parslet/cause.rb:70:in `raise': Failed to match sequence ('-'? ('0' / [1-9] DIGIT{0, }) ('.' DIGIT{1, })? ([eE] ('+' / '-')? DIGIT{1, })?) at line 1 char 1. (Parslet::ParseFailed)
	from /Users/jfrey/.gem/ruby/3.1.5/gems/parslet-2.0.0/lib/parslet/atoms/base.rb:49:in `parse'
	from (irb):3:in `<main>'
	from bin/console:17:in `<main>'

File: spec/workflow/intrinsic_function/parser_spec.rb (hunk 5 above)
Comment (Fryguy) [reply spec/workflow/intrinsic_function/parser_spec.rb:163]: Yes.  That is true and so is a bug.  I was thinking it was bad input, and forgot about that case.

Comment (Fryguy) [reply spec/workflow/intrinsic_function/parser_spec.rb:163]: Pointing back to https://github.com/ManageIQ/floe/pull/194#discussion_r1662606992, I'm going to add using Floe::Workflow::Path in this PR, because it will handle those cases and I can add support for context at the same time automatically.

Comment (agrare) [reply spec/workflow/intrinsic_function/parser_spec.rb:163]: :+1:

File: lib/floe/workflow/intrinsic_function/parser.rb [hunk 7]
```diff
@@ -0,0 +1,65 @@
...
+        rule(:states_array) do
+          (
+            str("States.Array") >> str('(') >> args >> str(')')
+          ).as(:states_array)
+        end
+
+        rule(:states_uuid) do
+          (
+            str("States.UUID()")
+          ).as(:states_uuid)
```
agrare, line 57: Rubocop has a good point could this just be
```suggestion
          str("States.UUID()").as(:states_uuid)
```

spec/workflow/intrinsic_function/parser_spec.rb
Fryguy: Ok done.  That came out nicer than I expected - I didn't realize Floe::Workflow::Path was so nicely self-contained.

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -0,0 +1,27 @@
...
+  class Workflow
+    module IntrinsicFunction
+      class Transformer < Parslet::Transform
+        rule(:null_literal  => simple(:v)) { nil }
+        rule(:true_literal  => simple(:v)) { true }
+        rule(:false_literal => simple(:v)) { false }
+
+        rule(:string   => simple(:v)) { v.to_s }
+        rule(:number   => simple(:v)) { v.match(/[eE.]/) ? Float(v) : Integer(v) }
+        rule(:jsonpath => simple(:v)) { Floe::Workflow::Path.value(v.to_s, context, input) }
```
agrare, line 18: Awesome

agrare:
Okay I think all that is left is if you want to use Floe.logger for debugging the tree or dropping it for now

Fryguy:
@agrare Updated to use Floe.logger. Example:

```ruby
$ bin/console
irb(main):001:0> Floe.logger.level = 0
=> 0
irb(main):002:0> Floe::Workflow::IntrinsicFunction.evaluate("States.Array('foo', 1, 1.0, true, false, null, $.input)", {}, {"input" => {"foo" => "bar"}})
D, [2024-07-02T10:59:07.361965 #55278] DEBUG -- : Parsed intrinsic function: "States.Array('foo', 1, 1.0, true, false, null, $.input)" => {:states_array=>[{:string=>"foo"@14}, {:number=>"1"@20}, {:number=>"1.0"@23}, {:true_literal=>"true"@28}, {:false_literal=>"false"@34}, {:null_literal=>"null"@41}, {:jsonpath=>"$.input"@47}]}
=> ["foo", 1, 1.0, true, false, nil, {"foo"=>"bar"}]
```

agrare (approved):

Fryguy:
@agrare One other thing I noticed is that in a failed parse, it will raise a `Parslet::ParseFailed` error.  Do you want me to trap and log that and/or wrap in a Floe-specific exception?  Or can I wait for a separate PR for that?

agrare:
Oh yeah, if it is an invalid payload definition we can raise a `Floe::InvalidWorkflowError` error if it is an invalid arg or something like that we can return a `States.IntrinsicFailure` error payload

kbrock:
Is there a way to ask if this is a valid string at initial parsing time?

agrare:
Yeah we were just discussing how ideally we'd separate "is the payload definition valid" vs "was it a runtime / argument error"

Fryguy:
> Is there a way to ask if this is a valid string at initial parsing time?

It depends how complicated we want to make this, but the parser and the transformer are separated, so that could be where the logical syntax/semantics boundary lives.  In my mind that's where the separation of `Floe::InvalidWorkflowError` and `States.IntrinsicFailure`.

However, a good example of where the lines get blurred is that States.UUID is defined to not take any parameters, so if someone passes `States.UUID(1)` is that a syntax error or a semantic error?  Right now it's a syntax error, but if we want it to be a semantic error, we'd have to changed the syntax definition in that it can take parameters, but then later fail with a semantic error if any parameters are given. 🤷

PR #197, opened 2024-05-29
Drop step_nonblock_wait

kbrock:
reduce the number of "run" methods

goal lead to another goal / thread
- simplify workflow so parallel and map are easier to implement
- reducing the manipulation of context in workflow
- reduce the number of run/step methods in workflow
- reduce the number of run/step methods in state

### DIFF
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -126,10 +126,6 @@ module Floe
       current_state.run_nonblock!
     end
 
-    def step_nonblock_wait(timeout: nil)
-      current_state.wait(:timeout => timeout)
-    end
-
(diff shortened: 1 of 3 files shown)


agrare:
@kbrock as long as Parallel implements ready? I don't think wait makes implementing Parallel any harder, but I could be wrong.

kbrock:
@agrare I noticed that this is not called by provider, exe/floe, or the test suite.
The only use case I could see was in tests. Since that was missing, I thought it better to remove code that is not used.

Agreed, parallel/map probably works great.

I was thinking dropping would make the change to pass in context on run easier.

PR #198, opened 2024-05-29
Drop context state machine and task

kbrock:
any time I add funcitonality to context, my PRs show as red.
It says there are too many methods on context.

This is an attempt to remove a few unused methods from context

task holds an id for a human interaction task.


agrare:
These are here to match the AWS Step Functions context as closely as possible so that an ASL written for step-functions would work with floe.

https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html

kbrock:
I had seen why we added them, but they are just there and not used. Was thinking we'd remove them until we started implementing them.

I still feel these are not needed but ... what are you going to do

Closing

PR #199, opened 2024-05-31
ResultPath=$ replaces complete output

kbrock:
ok, https://states-language.net/#error-output

> The default value, if the "ResultPath" field is not provided, is "$", meaning that the output consists entirely of the Error Output.

Which differs from the previous code
 https://github.com/ManageIQ/floe/blob/master/lib/floe/workflow/reference_path.rb#L28-L29

> If the payload is '$' then merge the value into the context

This fixes issues where the output is a simple class inserting at the top level

I stumbled across this because the `Task` `Catch` should output the error as the output node

### DIFF
--- a/lib/floe/workflow/reference_path.rb
+++ b/lib/floe/workflow/reference_path.rb
@@ -25,12 +25,9 @@ module Floe
       def set(context, value)
         result = context.dup
 
-        # If the payload is '$' then merge the value into the context
-        # otherwise store the value under the path
-        #
-        # TODO: how to handle non-hash values, raise error if path=$ and value not a hash?
+        # If the payload is '$' then replace the output with the value
         if path.empty?
-          result.merge!(value)
+          result = value.dup
         else
           child    = result
           keys     = path.dup
--- a/spec/workflow/reference_path_spec.rb
+++ b/spec/workflow/reference_path_spec.rb
@@ -50,7 +50,7 @@ RSpec.describe Floe::Workflow::ReferencePath do
 
   describe "#set" do
     let(:payload) { "$" }
-    let(:input) { {} }
+    let(:input) { {"old" => "key"} }
 
     context "with a simple path" do
       it "sets the output at the top-level" do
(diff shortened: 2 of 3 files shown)


kbrock:
update:
- ugh. fixed last minute optimization (that flubbed things up)

lib/floe/workflow/reference_path.rb
```diff
@@ -23,27 +23,15 @@ def get(context)
...
-          keys.each do |key|
-            child[key] = {} if child[key].nil?
-            child = child[key]
-          end
-
-          child[last_key] = value
+        return value.dup if path.empty?
+
+        child = result = context.dup
+        path[..-2].each do |key|
```
Fryguy, line 29: `path[..-2]` is melting my brain...is it chomping off the last character?  If so, is it chomping off a specific character (asking cause then you can use the `chomp` method with a parameter.

kbrock: yea. I wanted to get rid of the dup, but maybe this is the same.

This is
```diff
- keys     = path.dup
- last_key = keys.pop
- keys.each do |key|
+ path[..-2].each do
  ...
  end
- child[last_key]
+ child[path.last]
```

probably same as:
```ruby
- keys     = path.dup
- last_key = keys.pop
+ *keys, last_key = path
```

Though all of these may allocate the same so this may be an unnecessary change.

lib/floe/workflow/reference_path.rb
```diff
@@ -23,27 +23,15 @@ def get(context)
...
 
       def set(context, value)
-        result = context.dup
-
-        # If the payload is '$' then merge the value into the context
-        # otherwise store the value under the path
-        #
-        # TODO: how to handle non-hash values, raise error if path=$ and value not a hash?
-        if path.empty?
-          result.merge!(value)
```
kbrock, line 33: This is the main change

```diff
  if path.empty?
-   result.merge!(value)
+   value.dup
```

spec/workflow/reference_path_spec.rb
```diff
@@ -50,19 +50,27 @@
 
   describe "#set" do
     let(:payload) { "$" }
-    let(:input) { {} }
+    let(:input) { { "old" => "key"} }
```
kbrock, line 53: this was an empty hash before, so the tests did not show the distinction here.
Adding a before value caused the code to show the nuance I am highlighting

agrare:
Great catch @kbrock

lib/floe/workflow/reference_path.rb
agrare: @kbrock can we tackle this in a different refactoring PR if https://github.com/ManageIQ/floe/pull/199/files#r1622463119 is the primary fix here?

spec/workflow/reference_path_spec.rb
agrare, line 53: Rubocop has a good point
```suggestion
    let(:input) { {"old" => "key"} }
```

kbrock:
update:
- rolled back code changes and kept value replace
- fixed cop with spacing

lib/floe/workflow/reference_path.rb
kbrock: I don't know if it is any better.
Will probably forget all about this change. (as it should be)

Good call getting in the important stuff and changing as little as possible

agrare (approved):

PR #203, opened 2024-06-01
Pass credentials around with context

kbrock:
Overview
========

My higher goal is to have `State` not store the `workspace`, `context`, nor `credentials`. See #210 

We store `Context` and `Credentials` separately in the database. So it makes sense to keep them separate.

But having a single container that holds all runtime state simplifies matters.

Before
======

Passed `Context` and `Credentials` separately.
The `State` accessed them using `State#workspace#context` and `State#workspace#credentials`.

If we go towards removing `State#workspace`, `run_nonblock(context, credentials)` will provide that data.

After
=====

`Context` contains the context and `Context#to_h` returns the same value. But `Context#credentials` will contain the sensitive credentials.

The `State` accesses them using `State#workspace#context` and `State#workspace#context#credentials.

If we go towards removing `State#workspace`, `run_nonblock(context)` will be able to provide that data.

Compromise
==========

Would have liked to remove `Workspace#context` and `Workspace#initialize(..., context`), but left it in there until we update `manageiq-providers-workspace`.

### DIFF
--- a/exe/floe
+++ b/exe/floe
@@ -57,8 +57,8 @@ credentials =
 
 workflows =
   args.each_slice(2).map do |workflow, input|
-    context = Floe::Workflow::Context.new(opts[:context], :input => input)
-    Floe::Workflow.load(workflow, context, credentials)
+    context = Floe::Workflow::Context.new(opts[:context], :input => input, :credentials => credentials)
+    Floe::Workflow.load(workflow, context)
   end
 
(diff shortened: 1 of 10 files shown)


kbrock:
update:
- rebased

kbrock:
update:
- rebase

agrare (approved):

PR #204, opened 2024-06-01
Simpler State#long_name

kbrock:
Since the class name was derived from the type, just use the type variable

(this also makes it easier to output this information if we were logging without the class instantiated)

Logs have the same before/after


kbrock:
update:
- rebase

kbrock:
update:
- rebase

agrare (approved):

PR #206, opened 2024-06-06
State only modifies Context#state - prep for Map/Parallel

kbrock:
In prep for `Map`, have State only modify its own data in `Context`. (Followup will pass `context` into `State` methods, so the caller (i.e.: `Map`, `Parallel`, or `Workspace`) will determine the context and act accordingly.

1. Don't modify the context in `Workflow.new` - instead modify it in the first step.
2. `State#start` only modifies `Context#state` and not `Context#execute` - instead `Workflow` is responsible for `execute`.
3. `State#finish` only modified `Context#state` and not `Context#execute` - instead `Workflow` is responsible for that.

### DIFF
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -105,20 +105,17 @@ module Floe
 
       @states         = payload["States"].to_a.map { |state_name, state| State.build!(self, state_name, state) }
       @states_by_name = @states.each_with_object({}) { |state, result| result[state.name] = state }
-
-      unless context.state.key?("Name")
-        context.state["Name"] = start_at
-        context.state["Input"] = context.execution["Input"].dup
-      end
     rescue => err
       raise Floe::InvalidWorkflowError, err.message
     end
 
     def run_nonblock
+      start_workflow
       loop while step_nonblock == 0 && !end?
       self
     end
 
+    # NOTE: If running manually, make sure to call start_workflow at startup
     def step_nonblock
       return Errno::EPERM if end?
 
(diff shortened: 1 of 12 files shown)


kbrock:
update:
- fixed cops
- added comment
- indentation

lib/floe/workflow.rb
```diff
@@ -124,6 +119,12 @@ def step_nonblock
 
       step_next
       current_state.run_nonblock!
+                   .tap do
+        # Avoiding State#running? because that is potentially expensive.
+        # State#run_nonblock! already called running? via State#ready? and
+        # called State#finished -- which is what Context#state_finished? is detecting
+        context.execution["EndTime"] = context.state["FinishedTime"] if context.state_finished? && context.next_state.nil?
```
agrare, line 126: The .tap without actually using the return looks odd, could we just do
```ruby
result = current_state.run_nonblock!
context.execution["EndTime"] = context.state["FinishedTime"] if context.state_finished? && context.next_state.nil?
result
```

lib/floe/workflow.rb
```diff
@@ -155,6 +156,11 @@ def end?
     end
 
     def current_state
+      unless context.state.key?("Name")
+        context.state["Name"] = start_at
+        context.state["Input"] = context.execution["Input"].dup
+        context.execution["StartTime"] = Time.now.utc.iso8601
+      end
```
agrare, line 163: I think asking for the current_state potentially setting it is a bit overloaded, what if we had a `start_workflow!` method which set the workflow execution["StartTime"] and set up the initial state?  Then just either handle if `current_state.nil?` or call `start_workflow!` before any checks to `current_state`

lib/floe/workflow.rb
agrare: And if we have a `start_workflow!` [[Ref]](https://github.com/ManageIQ/floe/pull/206/files#r1631302239) we could call this `end_workflow!` or `finish_workflow!`

kbrock: really like this

lib/floe/workflow.rb
kbrock: made these changes.
you can see that `current_state` which is called a lot in specs needed the `start_workflow` - so a bunch of changes

I created a commit before these changes that reworked tests to have less stubbing and possibly call current_state less. This reduced in the test changes.

kbrock:
update:
- created method for start_workflow
- moved start_workflow out of current_state - this caused a lot of spec churn

update:
- oops, added end_workflow
- cops

kbrock:
@agrare looking forward to seeing #208 leveraging this

agrare (approved):

PR #209, opened 2024-06-07
[WIP] Common error message helper.

kbrock:
## Overview.

This is the 3rd attempt.

Depends upon:

- [ ] #229 

Goals

- Provide more verbose error messages.
- Introduce more pedantic error checking.
- Move more error checking to initialize time.

#### Parsing

- Ensure required fields are defined (see note).
- Reject fields that are the wrong type.
- Reject unknown fields.

#### Next processing

- Ensure `Next`/`Default` are valid.
- Add `Next` validation to `ChoiceRule` and `Catcher`.
- Reject a invalid `Next` field in a child `ChoiceRule`.
- Remove `State` access of `Workflow#payload["States"]` for validations (punted).

#### Reduction of scope

I did the following changes to reduce the scope of this

- No longer type checking/requiring on every field (booleans, timestamps, strings).
- No longer rejecting unknown fields.
- Fixes to Choice have been extracted to #189

I can easily add the first 2 items back but I was trying to keep this down.

## Before / After

```diff
- Missing field "States"
+ Workflow requires Hash field "States"
- Missing field "StartAt"
+ Workflow requires String field "StartAt"
- "StartAt" not in the "States" field
+ Workflow requires field "StartAt" to be in "States" list but got [Foo]
- Choice state must have "Choices"
+ States.Choice1 requires Array field "Choices"
- "Choices" must be a non-empty array
+ States.Choice1 requires Array field "Choices"
- "Default" not in "States"
+ States.Choice1 requires field "Default" to be in "States" list but got [MissingState]
- Invalid state type: [Invalid]
+ States.FirstState requires String field "Type" but got invalid value [Invalid]
- Missing "Type" field in state [FirstState]
+ States.FirstState requires String field "Type"
- (ruby runtime error)
+ States.FirstMatchState.Choices.0.Data requires Path field "Variable"
- Path [wrong] must start with "$"
+ States.FirstMatchState.Choices.0.Data requires Path field "Variable" Path [wrong] must start with "$"
- (ignored)
+ States.FirstMatchState.Choices.0.Not.0.Data does not recognize field "Next"
```

## Changelog History

1. Pass a Validator instead of `workflow`.
2. Pass an Enhance payload with getter methods.
3. `State#include Validation`. **<==**

Have another branch that:
- Adds booleans, timestamps, and strings back.
- Rejects unknown fields
- uses a dsl

### DIFF
--- a/lib/floe/workflow/states/wait.rb
+++ b/lib/floe/workflow/states/wait.rb
@@ -8,23 +8,18 @@ module Floe
       class Wait < Floe::Workflow::State
         include NonTerminalMixin
 
-        attr_reader :end, :input_path, :next, :seconds, :seconds_path, :timestamp, :timestamp_path, :output_path
+        fields do
+          state_ref "Next"
+          boolean "End"
+          number "Seconds"
+          timestamp "Timestamp"
+          path "TimestampPath"
+          path "SecondsPath"
+          path "InputPath", :default => "$"
+          path "OutputPath", :default => "$"
 
-        def initialize(workflow, name, payload)
-          super
-
-          @next           = state_ref!("Next", payload["Next"], workflow)
-          @end            = boolean!("End", payload["End"])
-          @seconds        = payload["Seconds"]&.to_i
-          @timestamp      = payload["Timestamp"]
-          @timestamp_path = path!("TimestampPath", payload["TimestampPath"])
-          @seconds_path   = path!("SecondsPath", payload["SecondsPath"])
-
-          @input_path     = path!("InputPath", payload.fetch("InputPath", "$"))
-          @output_path    = path!("OutputPath", payload.fetch("OutputPath", "$"))
-
-          require_fields!("Next" => @next, "End" => @end)
-          require_fields!("Seconds" => @seconds, "Timestamp" => @timestamp, "TimestampPath" => @timestamp_path, "SecondsPath" => @seconds_path)
+          require_set "Next", "End"
+          require_set "Seconds", "Timestamp", "TimestampPath", "SecondsPath"
         end
 
         def start(context)
--- a/lib/floe/workflow/states/wait.rb
+++ b/lib/floe/workflow/states/wait.rb
@@ -17,11 +17,11 @@ module Floe
           @end            = !!payload["End"]
           @seconds        = payload["Seconds"]&.to_i
           @timestamp      = payload["Timestamp"]
-          @timestamp_path = Path.new(payload["TimestampPath"]) if payload.key?("TimestampPath")
-          @seconds_path   = Path.new(payload["SecondsPath"]) if payload.key?("SecondsPath")
+          @timestamp_path = path!("TimestampPath", payload["TimestampPath"])
+          @seconds_path   = path!("SecondsPath", payload["SecondsPath"])
 
-          @input_path  = Path.new(payload.fetch("InputPath", "$"))
-          @output_path = Path.new(payload.fetch("OutputPath", "$"))
+          @input_path     = path!("InputPath", payload.fetch("InputPath", "$"))
+          @output_path    = path!("OutputPath", payload.fetch("OutputPath", "$"))
 
           validate_state!(workflow)
         end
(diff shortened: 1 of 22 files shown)


kbrock:
update:
- introduced common validation helper methods

update:
- rubocops

kbrock:
update:
- trying to reduce complexity

kbrock:
update:
- extra use of `validate_field`
- remove validation methods with a single line.

kbrock:
update:
- Introduce a validator (so we won't be passing around the workflow everywhere)
- NonTerminal now runs own validation and sets next/end

kbrock:
update:
- implemented common error validation for other PRs
- separate validator from workflow (will pass around)

update:
- rubocop fixes

update:
- rubocop fixes

kbrock:
update:
- no longer able to pass state into validator
- validator `state_for` no longer modifies selt.

update:
- remove no longer valid validator comments
- string interpolation

kbrock:
update:
- rebase (to help other PRs resolve conflicts)

lib/floe/workflow/states/wait.rb
```diff
@@ -8,22 +8,18 @@ module States
...
         include NonTerminalMixin
 
-        attr_reader :end, :input_path, :next, :seconds, :seconds_path, :timestamp, :timestamp_path, :output_path
+        attr_reader :input_path, :seconds, :seconds_path, :timestamp, :timestamp_path, :output_path
 
         def initialize(workflow, name, payload)
           super
 
-          @next           = payload["Next"]
-          @end            = !!payload["End"]
```
agrare, line 17: Why delete these two?  Per https://states-language.net/#state-type-table Wait requires either Next or End:true

agrare: Oh I see you moved these into NonTerminalMixin

kbrock: Good:
- everything dealing with Next/End is in one place
- No contract between the Mixin and the including class (currently, it assumes `@next` and `@end` are defined and set.
- Validation is no longer a contract between the two layers.

Bad:
- I really like the old way of having all variables explicitly stated in one place.

I guess I can change the way we parse it in `Wait#initialize` - that should fix it.

kbrock:
update:
- rebase
- no longer moving `End` and `Next` parsing in parent `NonTerminalMixin`
- validating Catch#`Next` is a valid ref, and `Catch#ErrorEquals` is a list
- validating `ChoiceRule` `Next` (first and second levels)
- validating `Retrier#ErrorEquals` is a list
- Fixed up Choice texts that were not valid

lib/floe/workflow/states/wait.rb
kbrock: Got rid of this change

kbrock:
update:
- dropped `key?` to make code climate happy

kbrock:
Goal is to simplify this PR

update:
- rebase
- build upon name_path
- dropped PayloadValidator that wrapped a payload, and instead have a mixin.
- don't handle strings, booleans, timestamps, nor detect missing fields.

kbrock:
update:
- trying to get the complexity of ChoiceRule#validate down

kbrock:
update:
- added reject_field to reduce complexity from 8 => 5

kbrock:
update:
- reducing changed line count `+232/-110` => `+218/-78`

kbrock:
update:
- rebased onto newest name_path
- rearrange a few methods in validation mixin methods
- now passing workspace to all objects that need validation
- fixed 2 specs that had Next and End
-

kbrock:
WIP: This keeps starting small and growing large. Conflating too many ideas. Going to move the path code into the full name change.

This may become multiple PRs:

- unify error messages
- common helpers to make validation easier
- validate all things

kbrock:
I'm merging the dsl work into here and I squashed the commits down to the various forms of this code (did not keep the payload validator.)

- handle paths and complicated parsing
- handle every field (with larger datatype coverage)
- dsl to "simplify" some of the data concepts (having the object know the fields does makes validation and parsing easier)

This has run its course with a good chunk of the error message changes being promoted to #229

PR #210, opened 2024-06-10
State does not store workflow or context

kbrock:
depends upon:
- [ ] #209 
- [x] #203

When you setup a `Parallel` state, a single `State` may be run with multiple `Context` values.

So this changes `State` to pass the `Context` on execute (so a single `State` can handle multiple flows. - a kind of command/flyweight pattern if you will.

This is split up into quite a few commits to try and make the progression more palatable.

### DIFF
--- a/lib/floe.rb
+++ b/lib/floe.rb
@@ -7,6 +7,7 @@ require_relative "floe/logging"
 
 require_relative "floe/runner"
 
+require_relative "floe/validator"
 require_relative "floe/workflow"
 require_relative "floe/workflow/catcher"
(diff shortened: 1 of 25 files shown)


kbrock:
update:
- depends upon common errors
- no longer passing `workflow` into `State.new`/`State.build` - passing `validator` instead

kbrock:
update:
- rebase

kbrock:
update:
- rebase (moved a bunch of this logic into parent #209 )

kbrock:
merged #216 instead

PR #211, opened 2024-06-10
Set StateHistory in Workflow not State

agrare:
NOTE I moved step_next after the run, since we have a dedicated start_workflow which sets the initial state context we no longer need step_next to pull double-duty here

### DIFF
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -120,16 +120,11 @@ module Floe
       return Errno::EPERM if end?
 
       result = current_state.run_nonblock!
+      return result if result != 0
 
       # if it completed the step
-      if result == 0
-        context.state_history << context.state
-        if context.next_state.nil?
-          end_workflow
-        else
-          step_next
-        end
-      end
+      context.state_history << context.state
+      context.next_state ? step! : end_workflow!
 
       result
     end
@@ -175,13 +170,6 @@ module Floe
       self
     end
 
-    # Avoiding State#running? because that is potentially expensive.
-    # State#run_nonblock! already called running? via State#ready? and
-    # called State#finished -- which is what Context#state_finished? is detecting
-    def end_workflow
-      context.execution["EndTime"] = context.state["FinishedTime"]
-    end
-
     # NOTE: Expecting the context to be initialized (via start_workflow) before this
     def current_state
       @states_by_name[context.state_name]
@@ -189,7 +177,7 @@ module Floe
 
     private
 
-    def step_next
+    def step!
       next_state = {"Name" => context.next_state}
 
       # if rerunning due to an error (and we are using Retry)
@@ -200,6 +188,13 @@ module Floe
       end
 
       context.state = next_state
+    end
+
+    # Avoiding State#running? because that is potentially expensive.
+    # State#run_nonblock! already called running? via State#ready? and
+    # called State#finished -- which is what Context#state_finished? is detecting
+    def end_workflow!
+      context.execution["EndTime"] = context.state["FinishedTime"]
     end
   end
 end
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -120,8 +120,11 @@ module Floe
       return Errno::EPERM if end?
 
       result = current_state.run_nonblock!
-      step_next
-      end_workflow
+
+      if context.state_finished?
+        step_next
+        end_workflow
+      end
 
       result
     end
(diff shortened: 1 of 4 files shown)


spec/workflow/states/task_spec.rb
```diff
@@ -264,21 +264,21 @@
             expect_run_async(input, :error => "States.Timeout")
             workflow.step_nonblock
 
-            expect(ctx.next_state).to          eq("State")
+            expect(ctx.state["Name"]).to       eq("State")
```
agrare, line 267: I would have liked not having to change any specs due to the first change moving where we step, but in this case I think the test was specific to where step_next was called rather than the actual runtime state.  The key here is that after running "step" the current state is the retried one and next_state isn't set yet.

kbrock: I agree with this change.
I am feeling the change to `step_next` is getting closer to our original run and makes running a single state the most complete.

I mentioned out of band `s/state["Name"]/state_name`

Thinking we can drop `next_state` from our tests and just test `state_name`. (another PR)

lib/floe/workflow.rb
```diff
@@ -119,8 +119,11 @@ def run_nonblock
     def step_nonblock
       return Errno::EPERM if end?
 
+      result = current_state.run_nonblock!
       step_next
-      current_state.run_nonblock!.tap { end_workflow }
+      end_workflow
```
kbrock, line 124: think we only want to `step_next ; end_workflow` if the state is successful (return 0)

kbrock: added a diff below showing the issue

kbrock:
```diff
diff --git a/lib/floe/workflow.rb b/lib/floe/workflow.rb
index 69b0709..26cb872 100644
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -120,8 +120,10 @@ module Floe
       return Errno::EPERM if end?

       result = current_state.run_nonblock!
-      step_next
-      end_workflow
+      if context.state_finished?
+        step_next
+        end_workflow
+      end

       result
     end
diff --git a/spec/workflow/states/wait_spec.rb b/spec/workflow/states/wait_spec.rb
index 0a17a85..e907d3d 100644
--- a/spec/workflow/states/wait_spec.rb
+++ b/spec/workflow/states/wait_spec.rb
@@ -90,6 +90,19 @@ RSpec.describe Floe::Workflow::States::Pass do
         end
         expect(state.running?).to be_falsey
       end
+
+      it "runs" do
+        workflow.run_nonblock
+        expect(workflow.end?).to eq(false)
+        expect(workflow.context.state_history.size).to eq(0)
+
+        Timecop.travel(Time.now.utc + 10) do
+          workflow.run_nonblock
+        end
+
+        expect(workflow.end?).to eq(true)
+        expect(workflow.context.state_history.size).to eq(2)
+      end
     end
```

lib/floe/workflow.rb
```diff
@@ -119,8 +119,14 @@ def run_nonblock
     def step_nonblock
       return Errno::EPERM if end?
 
-      step_next
-      current_state.run_nonblock!.tap { end_workflow }
+      result = current_state.run_nonblock!
+
+      if context.state_finished?
+        step_next
+        end_workflow
```
agrare, line 126: I wanted to drop the state_finished? from end_workflow but it is called in a bunch of specs

Fryguy (approved):

lib/floe/workflow.rb
```diff
@@ -164,30 +170,31 @@ def start_workflow
...
+      else
+        next_state["Input"] = context.output
       end
+
+      context.state = next_state
+    end
+
+    # Avoiding State#running? because that is potentially expensive.
+    # State#run_nonblock! already called running? via State#ready? and
+    # called State#finished -- which is what Context#state_finished? is detecting
```
kbrock, line 195: comment no longer valid.
we can remove later

lib/floe/workflow.rb
```diff
@@ -119,8 +119,14 @@ def run_nonblock
...
       return Errno::EPERM if end?
 
-      step_next
-      current_state.run_nonblock!.tap { end_workflow }
+      result = current_state.run_nonblock!
+      return result if result != 0
+
+      # if it completed the step
+      context.state_history << context.state
+      context.next_state ? step! : end_workflow!
```
kbrock, line 127: yes!

PR #212, opened 2024-06-12
[WIP] Move parameters into parent Mixins

kbrock:
Style question:

> Do we like the idea of moving the parameter reading into the mixin?

This shows what it looks like if we do this.

Depends upon:

- [x] #186 
- last 3 commits

Note: `Pass` does not support `"ResultSelector"`. The old code was ignoring it. This is ignoring it as well. Since `InputOutputMixin#process_output` was already referencing `result_selector`,
seemed ok to fully define that here.

---

WIP: Added WIP to mark this as a lower priority. it is ready

### DIFF
--- a/lib/floe/workflow/catcher.rb
+++ b/lib/floe/workflow/catcher.rb
@@ -5,7 +5,7 @@ module Floe
     class Catcher
       include Floe::Workflow::ErrorMatcherMixin
 
-      attr_reader :error_equals, :next, :result_path
+      attr_reader :next, :result_path
 
       def initialize(payload)
         @payload = payload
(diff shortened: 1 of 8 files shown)


kbrock:
update:
- rubocop indentation

kbrock:
WIP: this is a few commits in. not a conversation I want to have now
But I do like this code and would like to merge when we are ready to talk about it

kbrock:
update:
- fixed spec to properly display error (the payload was invalid in multiple ways - wrong error was returning)

kbrock:
update:
- cops

kbrock:
This does extract some complexity from various states and consolidates all the relevant parameters together.

But it just feels like a refactor right now and doesn't move us forward

PR #214, opened 2024-06-13
Handle non-json input/output values

agrare:
We currently assume that input/output have to be a JSON/hash payload, however while looking at some map examples it seems like it is possible for simple strings as well (and I assume by extension integers, etc...).

Example:
https://docs.aws.amazon.com/step-functions/latest/dg/tutorial-use-inline-map.html

Input to Map state:
```json
{
  "foo": "bar",
  "colors": [
    "red",
    "green",
    "blue",
    "yellow",
    "white"
  ]
}
```

So with `"ItemsPath": "$.colors"` the input to the `"ItemProcessor"` would be e.g. `"red"` and the output would be a simple UUID string.

### DIFF
--- a/lib/floe/workflow/context.rb
+++ b/lib/floe/workflow/context.rb
@@ -39,7 +39,7 @@ module Floe
       end
 
       def failed?
-        output&.key?("Error") || false
+        (output.kind_of?(Hash) && output.key?("Error")) || false
       end
 
       def ended?
--- a/lib/floe/workflow/state.rb
+++ b/lib/floe/workflow/state.rb
@@ -64,7 +64,7 @@ module Floe
         context.state["FinishedTime"] ||= finished_time.iso8601
         context.state["Duration"]       = finished_time - entered_time
 
-        level = context.output&.[]("Error") ? :error : :info
+        level = context.failed? ? :error : :info
         logger.public_send(level, "Running state: [#{long_name}] with input [#{context.input}]...Complete #{context.next_state ? "- next state [#{context.next_state}]" : "workflow -"} output: [#{context.output}]")
 
         0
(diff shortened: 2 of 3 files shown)


Fryguy:
Not sure In understand - it seems like the input to Map is an array always.  Or are you saying the contents of the Array itself can be anything (in this case an Array of Strings as opposed to an Array of Inputs/Hashes)?

agrare:
>  Or are you saying the contents of the Array itself can be anything

^ this

In the step-functions examples the input to Map is `["red", "green", et...]` which means the input to the sub-workflow is just `"red"`

agrare:
Also I think there isn't anything to prevent you from passing e.g. `"red"` as the top-level workflow input (if you don't have a Map state that is)

kbrock:
Can we solve this by removing the `JSON` parsing?
Instead we can put that parsing into `manageiq-workflow-providers`, `exe/floe`, and possibly in `Task`.

I've noticed that a lambda (task) can return a string (not json) and it will work. The basic guide stated some advantages of this. Not sure if we want to support this behavior. Think we throw away the input if it is not json.

agrare:
> Can we solve this by removing the JSON parsing?

Oh that's interesting, yeah that might be the way to go

kbrock:
I really like this PR

Fryguy:
Just wanted to add some clarifications here (as discussed with @agrare in PM).  Strings, integers, etc _are_ JSON.  I've updated the title to say non-hash (and Adam updated his commit before it was merged)

```
[1] pry(main)> require "json"
=> true
[2] pry(main)> "test".to_json
=> "\"test\""
[3] pry(main)> JSON.parse(_)
=> "test"
[4] pry(main)> 1.to_json
=> "1"
[5] pry(main)> JSON.parse(_)
=> 1
```

PR #215, opened 2024-06-17
Common Error messages alt

kbrock:
Alt to #209

This reads parameters, ensuring the data types, and no extra parameters.
The errors explain the exact location of the incorrect attribute.

Unlike 209, this wraps the params with an object that knows the location and has type checking.

Details
=======

- Fixes specs that don't define a `Next`.
- When state names are too long, display only the first 80 characters.
- All classes throws `InvalidWorkflowError` instead of relying upon later states to wrap unknown exceptions.
- Use `PayloadValidator` (instead of payload) to read values
- Detects unknown attributes.
- Enforces attribute types.
- All fields with validations. Separate PR for Choice, but now validates Paths
- Full context defined for `Path` or `ReferencePath` errors.

### DIFF
--- a/lib/floe.rb
+++ b/lib/floe.rb
@@ -7,6 +7,7 @@ require_relative "floe/logging"
 
 require_relative "floe/runner"
 
+require_relative "floe/payload_validator"
 require_relative "floe/workflow"
 require_relative "floe/workflow/catcher"
 require_relative "floe/workflow/choice_rule"
--- a/lib/floe/payload_validator.rb
+++ b/lib/floe/payload_validator.rb
@@ -21,17 +21,8 @@ module Floe
       @children    = children
     end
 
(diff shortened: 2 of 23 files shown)


kbrock:
update:
- cops

kbrock:
update
- code climate simplification (only 7 -> 6)

update:
- fixed cops from previous code climate change
- code climate simplification 2

kbrock:
update:
- introduce PayloadValidator#payload_template! to hopefully reduce code climate complexities and to make interface handle all examples

kbrock:
update
- dropped `fetch` and `key?` to drop code complexity
- ensured all paths and templates are using helper methods
-

kbrock:
merging this into #209 

Decided to go with this code base

PR #216, opened 2024-06-17
State without workflow

kbrock:
Alt to #210 but punting on the validation work done in #209 or #215

- This removes access to `Workflow`/`payload` inside the various `State` objects.
- That allows the `State` to be reused with different `input`.
- This passes these into the state validation, though that is more of a temporary fix.

### DIFF
--- a/lib/floe/workflow/states/choice.rb
+++ b/lib/floe/workflow/states/choice.rb
@@ -9,7 +9,7 @@ module Floe
         def initialize(workflow, name, payload)
           super
 
-          validate_state!
+          validate_state!(workflow)
 
           @choices = payload["Choices"].map { |choice| ChoiceRule.build(choice) }
           @default = payload["Default"]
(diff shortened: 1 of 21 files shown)


lib/floe/workflow/states/choice.rb
```diff
@@ -37,17 +37,17 @@ def end?
 
         private
 
-        def validate_state!
-          validate_state_choices!
-          validate_state_default!
+        def validate_state!(workflow, payload)
+          validate_state_choices!(payload)
+          validate_state_default!(workflow, payload)
```
agrare, line 42: Minor but isn't `@payload` set already here and we have an attr_reader?

I think you can drop this arg unless it is for another purpose

kbrock: changing.

But yea, I am not a fan of paylod or workflow.
(though it is pretty innocuous)

agrare (approved):

PR #219, opened 2024-06-18
tweaking code climate test limits

kbrock:
Possible Values can be found:
  https://docs.codeclimate.com/docs/default-analysis-configuration

A few limits I feel are off.
But this is here to start a conversation about what we want to modify.

- Drop `method-lines` and `file-lines` -- These issues would trigger a complexity issue.
- Drop `method-count` -- It does show complexity, so this is up to you all.
- Increase `complexity-logic` from 5 to 10.

For this repo, we have the following issues:

- `Workflow#initialize` complexity 17 > 5
- `Workflow#wait` complexity 34 > 5, ~~lines 45 > 25~~
- `Docker#wait` complexity 29 > 5, ~~lines 29 > 25~~
- `Kubernetes` methods 24 > 20, ~~lines 256 > 250~~
- `Kubernetes#initialize` complexity 13 > 5
- `Kubernetes#wait` complexity 21 > 5, ~~lines 36 > 25~~
- `Kubernetes#pod_spec` lines 40 > 25
- ~~`Kubernetes#cubeclient` complexity 9 > 5~~
- ~~`Context` methods 24 > 20~~
- `Context#initialize` complexity 14 > 5
- `Podman#global_docker_options` complexity 13 > 5
- ~~`Data#true?` lines 36 > 25~~
- ~~`ContainerRunner#resolve_cli_options!` complexity 7 > 5~~
- ~~`Task#initialize` complexity 7 > 5~~
- ~~`Task#parse_output` complexity 6 > 5~~
- ~~`State#wait` complexity 6 > 5~~

Striked out entries would go away.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -14,3 +14,13 @@ plugins:
     config: ".rubocop_cc.yml"
     channel: rubocop-1-56-3
 version: '2'
+checks:
+  complexity-logic:
+    config:
+      threshold: 10
+  method-count:
+    enabled: false
+  method-lines:
+    enabled: false
+  file-lines:
+    enabled: false


Fryguy:
I would think any changes here should really live in manageiq-style, unless you think they are floe-specific?

Fryguy:
@kbrock This morning in triage when looking at this we noticed that the manageiq-style .codeclimate.yml was never even installed.  @agrare opened a PR for that in https://github.com/ManageIQ/floe/pull/224.  I agree with tweaking the numbers, but think a) we should see if the manageiq-style .codeclimate.yml is sufficient and if not then I think it should be done over in manageiq-style if that makes sense globally.

kbrock:
Thanks.
Yes, lets use the default.

kbrock:
@Fryguy so I'm not seeing how these tests would apply here.

Can we add another download section that holds codeclimate specific values?

Fryguy:
> @Fryguy so I'm not seeing any parameters that are code climate related.

The parameters come from the rubocop definitions instead, I believe. So, for example, cyclomatic complexity is 11 here: https://github.com/ManageIQ/manageiq-style/blob/master/styles/cc_base.yml#L38.  I thought those values "overrode" the ones that are defaulted on code climate, but I could be wrong.

kbrock:
Yes, I see that cyclomatic complexity has been introduced. Think it was not running at all before.

The file limits (code climate) are still being enforced (you can see that this PR resolves a number of code climate issues - suggesting that those values are just for the https://ruby.sadi.st/Heckle.html tools, and max code length are for other code tools.

kbrock:
good enough for now

PR #220, opened 2024-06-18
Move the guts of the CLI into a class for easier testing

### LINKED ISSUE #218: [DEV] Improve performance of exe/floe specs
Overview
========

The `exe/floe` tests take 3 seconds while the rest of the suite takes 0.8s

Is there a way to rewrite these 12 examples to not take so long?

Numbers
=======

```
floe $ time bundle exec rake
Finished in 2.69 seconds (files took 0.12459 seconds to load)
403 examples, 0 failures

real	0m3.266s
user	0m2.156s
sys	0m0.800s

floe $ mv spec/exe/floe_spec.rb spec/exe/floe_spec-disabled.rb
floe $ time bundle exec rake
Finished in 0.26167 seconds (files took 0.1256 seconds to load)
391 examples, 0 failures

real	0m0.829s
user	0m0.592s
sys	0m0.179s
```

Fryguy:
```
Before: bundle exec rake  1.95s user 0.87s system 90% cpu 3.097 total
After:  bundle exec rake  0.67s user 0.25s system 94% cpu 0.981 total
```

I kept a singular test in there that ensures we are calling the exe/floe interface properly, but I'm really not sure if we need it or not.  I'd be ok with dropping it.  It would save an additional 0.13s knocking the time down to 0.85s or so

@kbrock Please review.
Fixes #218


lib/floe/cli.rb
```diff
@@ -0,0 +1,84 @@
...
+        Floe::ContainerRunner.cli_options(self)
+
+        banner("")
+        banner("General options:")
+      end
+
+      # Create workflow/input pairs from the various combinations of paramaters
+      workflows_inputs =
+        if opts[:workflow_given]
+          Optimist.die("cannot specify both --workflow and bare workflows") if args.any?
```
Fryguy, line 66: Now that this is in a class, these Optimist.die statements actually raise a SystemExit, which means this is kind of gross.  I started refactoring to use an Optimist::Parser, but then this code started getting too complicated, so left the PR as is to be a purer moving of the code.  We can follow up with better cleanup later if needed (or not - it's a CLI class after all).

kbrock: Think I added `raise SystemExit` so we could test for this case (rather than using `exit`)

PR #223, opened 2024-06-18
Separate workflow/context with different loggers

kbrock:
## Goal

The logging is global to all of floe. This means we do not associate certain logging information with each user/run.

Setting up a per execution log allows us to isolate the log per running instance and display it to the end user.

## Implementation

The `Context` is a per execution concept. It contains the user's information (in execution), the user's credentials, and the runtime context. (including credentials in context has conflated the db context and the runtime context)

Since we want a per execution logger, I added logger to the `Context`.

## Alternative

The `Workflow` currently contains the states and the runtime `Context`. The suggestion was to add the `logger` to the `Workflow`. Unfortunately, this does not work because while the various components have access to the `Context`, they do not have access to the `Workflow`.

So I propose we stay the course and use `Context#logger`

## Extracted

pulled out the following PRs:

- [ ] https://github.com/ManageIQ/floe/pull/289

### DIFF
--- a/lib/floe/cli.rb
+++ b/lib/floe/cli.rb
@@ -22,8 +22,7 @@ module Floe
 
       workflows =
         workflows_inputs.each_slice(2).map do |workflow, input|
-          context = Floe::Workflow::Context.new(opts[:context], :input => input, :credentials => credentials)
-          Floe::Workflow.load(workflow, context)
+          create_workflow(workflow, opts[:context], input, credentials)
         end
 
       Floe::Workflow.wait(workflows, &:run_nonblock)
@@ -81,6 +80,11 @@ module Floe
       Floe::ContainerRunner.resolve_cli_options!(opts)
 
       return workflows_inputs, opts
+    end
+
+    def create_workflow(workflow, context_payload, input, credentials)
+      context = Floe::Workflow::Context.new(context_payload, :input => input, :credentials => credentials)
+      Floe::Workflow.load(workflow, context)
     end
   end
 end
--- a/lib/floe/cli.rb
+++ b/lib/floe/cli.rb
@@ -23,6 +23,8 @@ module Floe
           create_workflow(workflow, opts[:context], input, credentials)
         end
 
+      output_streams = create_loggers(workflows, opts[:segment_output])
+
       logger.info("Checking #{workflows.count} workflows...")
       ready = Floe::Workflow.wait(workflows, &:run_nonblock)
       logger.info("Checking #{workflows.count} workflows...Complete - #{ready.count} ready")
@@ -35,6 +37,7 @@ module Floe
           logger.info("===")
         end
 
+        logger.info(output_streams[workflow].string) if output_streams[workflow]
         logger.info(workflow.output)
       end
 
@@ -59,6 +62,7 @@ module Floe
         opt :context, "JSON payload of the Context",              :type => :string
         opt :credentials, "JSON payload with Credentials",        :type => :string
         opt :credentials_file, "Path to a file with Credentials", :type => :string
+        opt :segment_output, "Segment output by each worker",     :default => false
 
         Floe::ContainerRunner.cli_options(self)
 
@@ -98,6 +102,18 @@ module Floe
     def create_workflow(workflow, context_payload, input, credentials)
       context = Floe::Workflow::Context.new(context_payload, :input => input, :credentials => credentials)
       Floe::Workflow.load(workflow, context)
+    end
+
+    def create_loggers(workflows, segment_output)
+      if workflows.size == 1 || !segment_output
+        # no extra work necessary
+        {}
+      else
+        workflows.each_with_object({}) do |workflow, h|
+          workflow.context.logger = Logger.new(output = StringIO.new)
+          h[workflow] = output
+        end
+      end
     end
   end
 end
--- a/lib/floe/cli.rb
+++ b/lib/floe/cli.rb
@@ -1,9 +1,10 @@
+require "floe"
+require "floe/container_runner"
+
 module Floe
   class CLI
     def initialize
       require "optimist"
-      require "floe"
-      require "floe/container_runner"
       require "logger"
 
       Floe.logger = Logger.new($stdout)
--- a/lib/floe/cli.rb
+++ b/lib/floe/cli.rb
@@ -3,6 +3,8 @@ require "floe/container_runner"
 
 module Floe
   class CLI
+    include Logging
+
     def initialize
       require "optimist"
       require "logger"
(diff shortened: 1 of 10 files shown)


kbrock:
update:
- rebased to handle changes to cli

think this is still part of discussions

Fryguy:
Overall I like the concept, but I think of context as more of the "serializable" part of the workflow, so it feels like the wrong place to put the logger.  I think workflow itself feels like the more appropriate place. @agrare @kbrock thoughts?

lib/floe/cli.rb
```diff
@@ -25,11 +26,25 @@ def run(args = ARGV)
...
         end
 
-      Floe::Workflow.wait(workflows, &:run_nonblock)
+      output_streams =
+        if workflows.size == 1 || !opts[:segment_output]
+          # no extra work necessary
+          {}
+        else
+          workflows.each_with_object({}) do |workflow, h|
+            workflow.context.logger = Logger.new(output = StringIO.new)
```
Fryguy, line 35: I'm not sure I like this.  I like seeing the live logs when using exe/floe, because I can see what state its in as it's running.

Fryguy: Oh wait, I misread the code.  I see that passing segment output will trigger this section otherwise it works normally. 👍

kbrock:
@Fryguy Yes, @agrare feels the same way as you do.

- `source.asl` -- source code.
- AST -- source code loaded into memory. This can be the same across multiple executions.
- Execution Environment
  - state input `Context#state.input` / `Context#input`
  - credentials `Context#credentials` (recent change)
  - execution history `Context#state_history`
  - logging (was suggesting `Context.logger` but up to you)

I was pushing that `Workflow` would only be the `AST`, but looks like we have linked `context` in there. I did try and pull `context` out, but it didn't work so well.

Maybe we want a different object to group `context` (really only need the current state), `credentials`, `logger`

kbrock:
update:
- rebase (conflict)

kbrock:
update:
- fix code climate (make Cli#run smaller)

update:
- fix refactor amuck.

kbrock:
update:
- rebase master

agrare:
Other than the logger belonging to Context I think this looks good, I would have thought logger lived with a Workflow

kbrock:
WIP: going to find another home for the logger. Possibly pass in a RuntimeContext, similar to an ImageProcessor context with Context State, credentials, and logger

kbrock:
un-WIP

Logistically, `Context#logger` works much better than `Workflow#logger`. The latter caused too many extra variables to get passed around, specifically in the `Runner`.

Philosophically, I feel the `Context` holds all execution and user specific state, while the `Workflow` holds no state and is valid across executions. By this definition, the logger belongs with the `Context`.

Elephant: Wihle I don't like the fact that `Workflow` currently has a reference to `Context`, I'm waiting for `wait` to settle down before suggesting any further change there.

lib/floe/cli.rb
```diff
@@ -82,5 +90,22 @@ def parse_options!(args)
...
+      Floe::Workflow.load(workflow, context)
+    end
+
+    def create_loggers(workflows, segment_output)
+      if workflows.size == 1 || !segment_output
+        # no extra work necessary
+        {}
+      else
+        workflows.each_with_object({}) do |workflow, h|
+          workflow.context.logger = Logger.new(output = StringIO.new)
```
Fryguy, line 105: Can you explain this string io part?  I don't understand this bit.  Is this meant to be like a NullLogger pattern (if so we should use an actual NullLogger like we do in other gems)?

lib/floe/cli.rb
```diff
@@ -18,19 +19,25 @@ def run(args = ARGV)
...
         workflows_inputs.each_slice(2).map do |workflow, input|
-          context = Floe::Workflow::Context.new(opts[:context], :input => input, :credentials => credentials)
-          Floe::Workflow.load(workflow, context)
+          create_workflow(workflow, opts[:context], input, credentials)
         end
 
-      Floe::Workflow.wait(workflows, &:run_nonblock)
+      output_streams = create_loggers(workflows, opts[:segment_output])
+
+      puts "Checking #{workflows.count} workflows..."
```
Fryguy, line 33: This won't be in the "real" application logs now though, right (since this is moved to cli, which is a dev tool)?  That was my concern.  Maybe it's not necessary in the application logs?

kbrock:
update:
- removed change to specs (no longer convert state.run to workflow.run)

lib/floe/cli.rb
kbrock: this is the cli. so yes, it will only go to stdout.
All entries in this file go to stdout via `puts`

Fryguy: I guess what I'm asking is before this change, this particular log line went to the application logs and now it doesn't.  I'm not sure how useful that is - maybe @agrare has opinions here.

agrare: Could we set the Floe.logger to stdout (if it isn't already) and
1. Keep the .wait logger the way it was
2. Use that instead of puts in cli.rb

lib/floe/workflow.rb
```diff
@@ -110,9 +108,12 @@ def run_nonblock
     def step_nonblock
       return Errno::EPERM if end?
 
+      start_state unless context.state_started?
       result = current_state.run_nonblock!(context)
       return result if result != 0
 
+      end_state
```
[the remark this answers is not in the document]
kbrock: Yea, `result != 0` means the state has not finished yet. Something like a `Wait` or `Task` that has not completed yet.

In the future, it will call into `run_nonblock`

But I think I'll take this out and revisit what I'm trying to do here. Thanks

lib/floe/cli.rb
kbrock: Most logging we have is feedback for individual workflows (mostly step start/complete)

The cli uses `State.wait` to orchestrate which workflows are run. I think of this as part of the command line interface. So going away from stdout seems strange to me. (asid: I do want to refactor this method to go even based, but that is a future action)

So is the thought is to convert all cli `puts` statements to log statements?

lib/floe/workflow.rb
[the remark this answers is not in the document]
Fryguy: oh i see it's the count of incomplete ones?  Interesting - Maybe result is just a bad variable name and something like "remaining_something" would be clearer.

kbrock:
update:
- output changed from `puts` to `logger`
- removed a number of extra changes
- drop start_state/end_state

update:
- added logging ()'s (code climate)

update:
- extra credentials to own method (code climate)

update:
- add logging ()'s (code climate - 1 remaining)

kbrock:
update:
- extracted cli refactoring to own PR
- rebased upon changes

Fryguy (approved):
LGTM - also want @agrare to review

kbrock:
@agrare You ok with this, or you want me to pull off one more PR and iterate a little more slowly.

I guess we could better partition runtime state and pull context out of there, but was hoping for that to be a future PR

agrare (approved):
I'm okay with this with the understanding that in the future we're going to split out our runtime context (lowercase c) from the Context :tm: object that is presented to the state machine

PR #224, opened 2024-06-19
Add a codeclimate config file

agrare:
Tell codeclimate to use our local rubocop config to pull in manageiq-style rules.


agrare:
I used github.com/ManageIQ/inventory_refresh as the template for this

Fryguy:
Strange this one is so different from the one in manageiq-style itself (https://github.com/ManageIQ/manageiq-style/blob/master/.codeclimate.yml).  I can merge anyway, but I'm wondering if we need more consistency there (just not sure in which direction)

agrare:
Oh I didn't realize that was a template for other repos, that looks good we can start with that and add back in anything else we end up needing

PR #225, opened 2024-06-19
Fix Input/Output handling for Pass, Choice, Succeed states

agrare:
Introduced by https://github.com/ManageIQ/floe/pull/191/files#diff-04416988853398624310337389715edeb1fed6c963a0a2a41ef882df8166dd6e

We're missing proper input processing for the Pass state if there is no Result attribute:

```
$ aws stepfunctions --endpoint-url http://localhost:8083 create-state-machine --name Pass --role-arn "arn:aws:iam::012345678901:role/DummyRole" --definition "{\"Comment\": \"Using Pass State\", \"StartAt\": \"Pass\", \"States\": {\"Pass\": {\"Type\": \"Pass\", \"End\": true, \"InputPath\": \"$.colors\"}}}"
$ aws stepfunctions --endpoint-url http://localhost:8083 start-execution --state-machine-arn arn:aws:states:us-east-1:123456789012:stateMachine:Pass --input '{"colors": "red"}'
$ aws stepfunctions --endpoint-url http://localhost:8083 describe-execution --execution-arn arn:aws:states:us-east-1:123456789012:execution:Pass:4747e8dd-c3ab-4124-bb45-1d7337018ddb
{
    "executionArn": "arn:aws:states:us-east-1:123456789012:execution:Pass:4747e8dd-c3ab-4124-bb45-1d7337018ddb",
    "stateMachineArn": "arn:aws:states:us-east-1:123456789012:stateMachine:Pass",
    "name": "4747e8dd-c3ab-4124-bb45-1d7337018ddb",
    "status": "SUCCEEDED",
    "startDate": "2024-06-19T12:59:05.702000-04:00",
    "stopDate": "2024-06-19T12:59:05.706000-04:00",
    "input": "{\"colors\": \"red\"}",
    "inputDetails": {
        "included": true
    },
    "output": "\"red\"",
    "outputDetails": {
        "included": true
    }
}
```

We should be getting `"red"` as the output, but currently we return the entire input payload:
```
Failures:

  1) Floe::Workflow::States::Pass#run_nonblock! Without Results Uses input
     Failure/Error: expect(ctx.output).to eq("red")
     
       expected: "red"
            got: {"color"=>"red"}
     
       (compared using ==)
     
       Diff:
       @@ -1 +1 @@
       -"red"
       +"color" => "red",
       
     # ./spec/workflow/states/pass_spec.rb:101:in `block (4 levels) in <top (required)>'
```

### DIFF
--- a/lib/floe/workflow/states/pass.rb
+++ b/lib/floe/workflow/states/pass.rb
@@ -25,7 +25,8 @@ module Floe
         end
 
         def finish(context)
-          context.output = process_output(context, result)
+          input = result.nil? ? process_input(context) : result
+          context.output = process_output(context, input)
           super
         end
 
--- a/spec/workflow/states/pass_spec.rb
+++ b/spec/workflow/states/pass_spec.rb
@@ -121,6 +121,16 @@ RSpec.describe Floe::Workflow::States::Pass do
           expect(ctx.output).to eq("red")
         end
       end
+
+      context "with InputPath, ResultPath, and OutputPath" do
+        let(:input)   { {"color" => "red", "garbage" => nil} }
+        let(:payload) { {"Pass" => {"Type" => "Pass", "End" => true, "InputPath" => "$.color", "ResultPath" => "$.results", "OutputPath" => "$.results"}} }
+
+        it "Uses OutputPath to drop other keys" do
+          workflow.run_nonblock
+          expect(ctx.output).to eq("red")
+        end
+      end
     end
   end
 end
--- a/spec/workflow/states/pass_spec.rb
+++ b/spec/workflow/states/pass_spec.rb
@@ -89,5 +89,38 @@ RSpec.describe Floe::Workflow::States::Pass do
         )
       end
     end
+
+    context "Without Results" do
+      let(:input) { {"color" => "red"} }
+      let(:payload) do
+        {"Pass" => {"Type" => "Pass", "End" => true}}
+      end
+
+      it "passes output through to input" do
+        workflow.run_nonblock
+        expect(ctx.output).to eq(input)
+      end
+
+      context "with InputPath" do
+        let(:payload) do
+          {"Pass" => {"Type" => "Pass", "End" => true, "InputPath" => "$.color"}}
+        end
+
+        it "Uses InputPath to select color" do
+          workflow.run_nonblock
+          expect(ctx.output).to eq("red")
+        end
+      end
+
+      context "with OutputPath" do
+        let(:input)   { {"color" => "red", "garbage" => nil} }
+        let(:payload) { {"Pass" => {"Type" => "Pass", "End" => true, "OutputPath" => "$.color"}} }
+
+        it "Uses OutputPath to drop other keys" do
+          workflow.run_nonblock
+          expect(ctx.output).to eq("red")
+        end
+      end
+    end
   end
 end
--- a/lib/floe/workflow/path.rb
+++ b/lib/floe/workflow/path.rb
@@ -26,8 +26,10 @@ module Floe
             [input, payload]
           end
 
-        results = JsonPath.on(obj, path)
+        # If path is $ then just return the entire input
+        return obj if path == "$"
 
+        results = JsonPath.on(obj, path)
         results.count < 2 ? results.first : results
       end
     end
--- a/lib/floe/workflow/states/choice.rb
+++ b/lib/floe/workflow/states/choice.rb
@@ -19,7 +19,8 @@ module Floe
         end
 
         def finish(context)
-          output     = output_path.value(context, context.input)
+          input      = input_path.value(context, context.input)
+          output     = output_path.value(context, input)
           next_state = choices.detect { |choice| choice.true?(context, output) }&.next || default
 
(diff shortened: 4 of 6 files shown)


kbrock:
I will put together a test with a combination of result, input path, result path, output path.

see how the values go.

I'm wondering if I screwed up and the input path IS processed for the output

Also noticed that Success has input/output path and we do not handle.

Fryguy (approved):

kbrock:
please merge #214 before merging this

lib/floe/workflow/states/pass.rb
```diff
@@ -25,7 +25,8 @@ def initialize(workflow, name, payload)
         end
 
         def finish(context)
-          context.output = process_output(context, result)
+          input = result.nil? ? process_input(context) : result
+          context.output = process_output(context, input)
```
kbrock, line 29: This makes all my pass tests pass.

But I think the fix belongs in `process_output`

```diff
        def process_output(context, results)
-          return context.input.dup if results.nil?
+          return process_input(context) if results.nil?
```

Not sure if we need a dup there or not.

agrare: I think doing `context.output = process_input(context) if results.nil?` would bypass the `OutputPath` part no?

agrare: I added a spec showing why this doesn't work with OutputPath, https://github.com/ManageIQ/floe/pull/225/files#diff-c4b20f6026d80247d4c5645a2811f66f0bd918ebb4763da5fb0d19c8a7d092faR115-R123

Fryguy:
Clarifying question - In the OP I see

> ```
> "output": "\"red\"",
> ```

but then the comment right after said

> We should be getting "red" as the output, but currently we return the entire input payload:

I think this is supposed to be that we should be getting `\"red\"` as the output (with the quotations), as that's valid JSON.  This overlaps with my concerns in https://github.com/ManageIQ/floe/pull/222#issuecomment-2182948069 as well.

spec/workflow/states/pass_spec.rb
```diff
@@ -89,5 +89,48 @@
...
+      end
+
+      context "with InputPath" do
+        let(:payload) do
+          {"Pass" => {"Type" => "Pass", "End" => true, "InputPath" => "$.color"}}
+        end
+
+        it "Uses InputPath to select color" do
+          workflow.run_nonblock
+          expect(ctx.output).to eq("red")
```
Fryguy, line 111: So I think this is wrong?  I think this should be

```suggestion
          expect(ctx.output).to eq("\"red\"")
```

agrare: Yeah I think you're right and that matches what the asl sim outputs from the description.

agrare: I think this is a slightly different issue than the one I'm trying to fix here though.  The core issue for this PR is we were completely ignoring InputPath, the issue you describe is around JSON and workflow/state input/output.  Think that is probably better solved in https://github.com/ManageIQ/floe/pull/222

kbrock: I feel all values should be ruby objects
And only the transitions to the outside world should convert json

So `CLI` and `Task` (converting their output to our `Result`) do conversion

lib/floe/workflow/states/pass.rb
kbrock: this did bypass output processing. So the solution would be:

```ruby
results ||= process_input(context
```

BUT, per an out of band conversation with @agrare:

> Let's not put input processing into output processing.

So the preferred implementation puts the fixing of `result.nil?` on the caller:

```ruby
process_output(context, results || process_input(context)
```

agrare: Is that just a one-line version of what I have ?

spec/workflow/states/pass_spec.rb
Fryguy: Yeha I was conflicted on whether context.output was in Ruby still or should already be converted to a JSON string.

agrare: @Fryguy I think https://github.com/ManageIQ/floe/pull/228/commits/b7377a89a6638bb1743cd35c174a12c2e57acbb0#diff-c4b20f6026d80247d4c5645a2811f66f0bd918ebb4763da5fb0d19c8a7d092fa covers your concern here, still very WIP but throw :tomato:

lib/floe/workflow/states/pass.rb
kbrock: Yes. that was from you. and I was trying to bring that conversation back here

spec/workflow/states/pass_spec.rb
agrare: Yeah I thought about it a bit more and I agree with @kbrock here, context is "internal" where workflow.output is defining the "user interface".  As long as the states get the right input/output and the workflow handles input/output properly everything is up to us.

I changed https://github.com/ManageIQ/floe/pull/228 to have workflow.output as JSON

Issue #221, opened 2024-06-18
Multiple simultaneous workflows are hard to tell apart in the logs


Fryguy:
See example below.

I think we can fix this by ensuring that log lines are prefixed with a GUID or a run identifier or some sorr

---

```
$ exe/floe examples/workflow.asl examples/workflow.asl --input '{"foo":1}'
I, [2024-06-18T16:16:16.300309 #72612]  INFO -- : checking 2 workflows...
I, [2024-06-18T16:16:16.300409 #72612]  INFO -- : Running state: [Task:FirstState] with input [{"foo"=>1}]...
D, [2024-06-18T16:16:16.302993 #72612] DEBUG -- : Running docker run --detach -e foo\=1 -e _CREDENTIALS\=/run/secrets -v /var/folders/47/l4t878ld45d2h11b45x88sr80000gn/T/20240618-72612-ulwuk9:/run/secrets:z --name floe-hello-world-c197ca50 docker.io/agrare/hello-world:latest
I, [2024-06-18T16:16:19.503519 #72612]  INFO -- : Running state: [Task:FirstState] with input [{"foo"=>1}]...Complete - next state [ChoiceState] output: [{"foo"=>1}]
I, [2024-06-18T16:16:19.573398 #72612]  INFO -- : Running state: [Choice:ChoiceState] with input [{"foo"=>1}]...
I, [2024-06-18T16:16:19.574053 #72612]  INFO -- : Running state: [Choice:ChoiceState] with input [{"foo"=>1}]...Complete - next state [FirstMatchState] output: [{"foo"=>1}]
I, [2024-06-18T16:16:19.574151 #72612]  INFO -- : Running state: [Task:FirstMatchState] with input [{"foo"=>1}]...
D, [2024-06-18T16:16:19.574398 #72612] DEBUG -- : Running docker run --detach -e foo\=1 --name floe-hello-world-26ef3e94 docker.io/agrare/hello-world:latest
I, [2024-06-18T16:16:22.204991 #72612]  INFO -- : Running state: [Task:FirstMatchState] with input [{"foo"=>1}]...Complete - next state [PassState] output: [{"foo"=>1}]
I, [2024-06-18T16:16:22.275456 #72612]  INFO -- : Running state: [Pass:PassState] with input [{"foo"=>1}]...
[...]

PR #226, opened 2024-06-20
Add an Execution unique ID to Context

agrare:
This will give each execution a unique reference that can be used for identifying a running workflow


agrare:
I thought this would pair well with https://github.com/ManageIQ/floe/pull/223 to help solve https://github.com/ManageIQ/floe/issues/221 since we could have the execution id as part of the logger

PR #227, opened 2024-06-21
Validate that state machine input is valid JSON

agrare:
When running a workflow validate that the state-machine input is valid JSON.  We were a little lax with specs passing Hashes in for convenience but it isn't hard to ensure we're passing JSON in to context input.

When running with ASL passing invalid input:
```
$ aws stepfunctions --endpoint-url http://localhost:8083 start-execution --state-machine-arn arn:aws:states:us-east-1:123456789012:stateMachine:Pass --input 'foo'

An error occurred (InvalidExecutionInput) when calling the StartExecution operation: Invalid State Machine Execution Input: 'Unrecognized token 'foo': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: (String)"foo"; line: 1, column: 4]'
```

With this PR:
```
$ exe/floe examples/workflow.asl 'foo'
Invalid State Machine Execution Input: unexpected token at 'foo': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
```

### DIFF
--- a/lib/floe/workflow/context.rb
+++ b/lib/floe/workflow/context.rb
@@ -9,9 +9,7 @@ module Floe
       # @param input [Hash] (default: {})
       def initialize(context = nil, input: nil, credentials: {})
         context = JSON.parse(context) if context.kind_of?(String)
-
-        input ||= {}
-        input = JSON.parse(input) if input.kind_of?(String)
+        input   = JSON.parse(input || "{}")
 
         @context = context || {}
         self["Execution"]          ||= {}
@@ -23,7 +21,7 @@ module Floe
 
         @credentials = credentials || {}
       rescue JSON::ParserError => err
-        raise Floe::InvalidWorkflowError, err.message
+        raise Floe::InvalidExecutionInput, "Invalid State Machine Execution Input: #{err}: was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')"
       end
 
       def execution
--- a/lib/floe.rb
+++ b/lib/floe.rb
@@ -42,6 +42,7 @@ require "time"
 module Floe
   class Error < StandardError; end
   class InvalidWorkflowError < Error; end
+  class InvalidExecutionInput < Error; end
 
   def self.logger
     @logger ||= NullLogger.new
--- a/lib/floe/cli.rb
+++ b/lib/floe/cli.rb
@@ -35,6 +35,8 @@ module Floe
       end
 
       workflows.all? { |workflow| workflow.context.success? }
+    rescue Floe::Error => err
+      abort(err.message)
     end
 
     private
--- a/lib/floe/workflow.rb
+++ b/lib/floe/workflow.rb
@@ -111,7 +111,7 @@ module Floe
       validate_workflow
 
       @states_by_name = @states.each_with_object({}) { |state, result| result[state.short_name] = state }
-    rescue Floe::InvalidWorkflowError
+    rescue Floe::Error
       raise
     rescue => err
       raise Floe::InvalidWorkflowError, err.message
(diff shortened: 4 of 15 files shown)


Fryguy:
I like how it matches the simulator :+1:

lib/floe/workflow/context.rb
```diff
@@ -9,9 +9,7 @@ class Context
       # @param input [Hash] (default: {})
       def initialize(context = nil, input: nil, credentials: {})
         context = JSON.parse(context) if context.kind_of?(String)
-
-        input ||= {}
-        input = JSON.parse(input) if input.kind_of?(String)
+        input   = JSON.parse(input || "{}")
```
kbrock, line 12: Like how the `if input...` has been removed.

Would prefer to change the default `input: "{}"` and drop the `input || "{}"`

agrare: the `input ||= {}` was if someone explicitly passed `:input => nil` it would still default to `{}`

kbrock: Heh. Yea.

I guess in general I want to keep the data closer to what the user provided.

agrare: I think the better question is "what does step functions do?", if I do
```
aws stepfunctions start-execution --input "null"
```

I get "null" output
```
{
    "executionArn": "arn:aws:states:us-east-1:123456789012:execution:Succeed:6cae58d1-1aa6-415f-a010-ddc730d6b81f",
    "stateMachineArn": "arn:aws:states:us-east-1:123456789012:stateMachine:Succeed",
    "name": "6cae58d1-1aa6-415f-a010-ddc730d6b81f",
    "status": "SUCCEEDED",
    "startDate": "2024-06-24T14:50:47.849000-04:00",
    "stopDate": "2024-06-24T14:50:47.876000-04:00",
    "input": "null",
    "inputDetails": {
        "included": true
    },
    "output": "null",
    "outputDetails": {
        "included": true
    }
}
```

So I think actually `{}` if not provided, but `nil` if explicitly passed `"null"` is actually correct.

kbrock: I think we are in total agreement here.

(I was just squabbling over who converted a `nil => {} / "null" => nil`, and was thinking it was the cli. But not my hill)

agrare: > (I was just squabbling over who converted a nil => {} / "null" => nil, and was thinking it was the cli. But not my hill)

Hey I'm open to why we shouldn't do it Context, but if we did it in cli then the same logic wouldn't apply to anyone using Context/Workflow as ruby classes right?

kbrock: > Hey I'm open to why we shouldn't do it Context, but if we did it in cli then the same logic wouldn't apply to anyone using Context/Workflow as ruby classes right?

Totally.
I double talk here. I feel that workflow is the public interface, but I want people to pass in a context.

Maybe we need a public interface for creating a context, like we have `Workflow#load`

Ok, the only thing I care about is when you are creating a Map, do you have a context for each parallel branch? Are you able to handle this conversion?

If so, lets :shipit:

agrare: > Ok, the only thing I care about is when you are creating a Map, do you have a context for each parallel branch? Are you able to handle this conversion?

Yes, each `ItemProcessor` iteration has its own `ItemProcessorContext` which lives under the top-level `Context["State"]` for the Map state.

agrare:
@Fryguy what do you think of this, something that we want to do?

Fryguy:
Yes I think this is something we want to do. Actually had a few questions of we can chat about it.

Fryguy (approved):

PR #229, opened 2024-06-21
Pass full path name to State.new for better errors

kbrock:
## Overview

`statelint` passes around the path to the states instead of just the name.

If we pass a full path to `State` and `ChoiceRule` we can improve validations

I want to go through each of the states to check all fields, but that just fixed up the existing errors and added a little bit of structure to help me fix ChoiceState

## Changes:

```diff
- State requires ErrorEquals
+ States.State.Retry.0 does not have required field "ErrorEquals"
- State requires ErrorEquals
+ States.State.Retry.0 does not have required field "ErrorEquals"
- Choice state must have "Choices"
+ States.Choice1 does not have required field "Choices"
- "Choices" must be a non-empty array
+ States.Choice1 field "Choices" must be a non-empty array
- "Choices" must be a non-empty array
+ States.Choice1 field "Choices" must be a non-empty array
- "Default" not in "States"
+ States.Choice1 field "Default" value "MissingState" is not found in "States"
- Missing field "States"
+ State Machine does not have required field "States"
- Invalid state type: [Invalid]
+ States.FirstState field "Type" value "Invalid" is not valid
- Missing field "StartAt"
+ State Machine does not have required field "StartAt"
- "StartAt" not in the "States" field
+ State Machine field "StartAt" value "Foo" is not found in "States"
- Missing "Type" field in state [FirstState]
+ States.FirstState does not have required field "Type"
- State name [#{truncated_state_name}] must be less than or equal to 80 characters
+ States field "Name" value "#{state_name}" must be less than or equal to 80 characters
- Invalid resource scheme [invalid]
+ States.FirstState field "Resource" value "invalid://foo" Invalid resource scheme [invalid]
```

### DIFF
--- a/lib/floe/validation_mixin.rb
+++ b/lib/floe/validation_mixin.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module Floe
+  module ValidationMixin
+    def self.included(base)
+      base.extend(ClassMethods)
+    end
+
+    def parser_error!(comment)
+      self.class.parser_error!(name, comment)
+    end
+
+    def missing_field_error!(field_name)
+      self.class.missing_field_error!(name, field_name)
+    end
+
+    def invalid_field_error!(field_name, field_value = nil, comment = nil)
+      self.class.invalid_field_error!(name, field_name, field_value, comment)
+    end
+
+    def workflow_state?(field_value, workflow)
+      workflow.payload["States"] ? workflow.payload["States"].include?(field_value) : true
+    end
+
+    def wrap_parser_error(field_name, field_value)
+      yield
+    rescue ArgumentError, InvalidWorkflowError => error
+      invalid_field_error!(field_name, field_value, error.message)
+    end
+
+    module ClassMethods
+      def parser_error!(name, comment)
+        name = name.join(".") if name.kind_of?(Array)
+        raise Floe::InvalidWorkflowError, "#{name} #{comment}"
+      end
+
+      def missing_field_error!(name, field_name)
+        parser_error!(name, "does not have required field \"#{field_name}\"")
+      end
+
+      def invalid_field_error!(name, field_name, field_value, comment)
+        # instead of displaying a large hash or array, just displaying the word Hash or Array
+        field_value = field_value.class if field_value.kind_of?(Hash) || field_value.kind_of?(Array)
+
+        parser_error!(name, "field \"#{field_name}\"#{" value \"#{field_value}\"" unless field_value.nil?} #{comment}")
+      end
+    end
+  end
+end
--- a/lib/floe/workflow/catcher.rb
+++ b/lib/floe/workflow/catcher.rb
@@ -5,10 +5,11 @@ module Floe
     class Catcher
       include Floe::Workflow::ErrorMatcherMixin
 
-      attr_reader :error_equals, :next, :result_path
+      attr_reader :error_equals, :next, :result_path, :name
 
-      def initialize(payload)
-        @payload = payload
+      def initialize(_workflow, name, payload)
+        @name         = name
+        @payload      = payload
 
         @error_equals = payload["ErrorEquals"]
         @next         = payload["Next"]
--- a/lib/floe/workflow/catcher.rb
+++ b/lib/floe/workflow/catcher.rb
@@ -3,7 +3,8 @@
 module Floe
   class Workflow
     class Catcher
-      include Floe::Workflow::ErrorMatcherMixin
+      include ErrorMatcherMixin
+      include ValidationMixin
 
       attr_reader :error_equals, :next, :result_path, :name
 
@@ -14,7 +15,8 @@ module Floe
         @error_equals = payload["ErrorEquals"]
         @next         = payload["Next"]
         @result_path  = ReferencePath.new(payload.fetch("ResultPath", "$"))
-        raise Floe::InvalidWorkflowError, "State requires ErrorEquals" if !@error_equals.kind_of?(Array) || @error_equals.empty?
+
+        missing_field_error!("ErrorEquals") if !@error_equals.kind_of?(Array) || @error_equals.empty?
       end
     end
   end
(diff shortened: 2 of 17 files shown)


kbrock:
update:
- using `_full_name` for Retrier. Won't use variable until we start validating those fields

kbrock:
update:
- rubocop

kbrock:
@agrare This doesn't do much, it just passes around the name. Now that I know this is a viable direction, I have updated #209 to include the before and afters

kbrock:
update:
- `s/Retrier/Catcher` -- a copy pasted error

lib/floe/workflow/states/task.rb
```diff
@@ -20,8 +20,8 @@ def initialize(workflow, name, payload)
           @resource          = payload["Resource"]
           @runner            = Floe::Runner.for_resource(@resource)
           @timeout_seconds   = payload["TimeoutSeconds"]
-          @retry             = payload["Retry"].to_a.map { |retrier| Retrier.new(retrier) }
-          @catch             = payload["Catch"].to_a.map { |catcher| Catcher.new(catcher) }
+          @retry             = payload["Retry"].to_a.each_with_index.map { |retrier, i| Retrier.new(full_name + ["Retry", i.to_s], retrier) }
+          @catch             = payload["Catch"].to_a.each_with_index.map { |catcher, i| Catcher.new(full_name + ["Catch", i.to_s], catcher) }
```
Fryguy, line 24:
```suggestion
          @retry             = payload["Retry"].to_a.each_with_index.map { |retrier, i| Retrier.new(full_name + ["Retry", i.to_s], retrier) }
          @catch             = payload["Catch"].to_a.each_with_index.map { |catcher, i| Catcher.new(full_name + ["Catch", i.to_s], catcher) }
```
```suggestion
          @retry             = payload["Retry"].to_a.map.with_index { |retrier, i| Retrier.new(full_name + ["Retry", i.to_s], retrier) }
          @catch             = payload["Catch"].to_a.map.with_index { |catcher, i| Catcher.new(full_name + ["Catch", i.to_s], catcher) }
```

kbrock:
update:
- Added the validation messages changes from #209
- Cut down the common validation method to just a hand full of messages
- Did not add any new messages, but rather tried to minimize change

update:
- reduced flog score to make code climate happy

lib/floe/workflow/catcher.rb
```diff
@@ -3,17 +3,19 @@
...
-        @payload = payload
+      def initialize(full_name, payload)
+        @full_name    = full_name
+        @payload      = payload
 
         @error_equals = payload["ErrorEquals"]
         @next         = payload["Next"]
         @result_path  = ReferencePath.new(payload.fetch("ResultPath", "$"))
-        raise Floe::InvalidWorkflowError, "State requires ErrorEquals" if !@error_equals.kind_of?(Array) || @error_equals.empty?
+        parse_error!("State requires ErrorEquals") if !@error_equals.kind_of?(Array) || @error_equals.empty?
```
kbrock, line 18: ugh

```suggestion
        parser_missing_field!("ErrorEquals") if !@error_equals.kind_of?(Array) || @error_equals.empty?

        
```
```suggestion
        parse_error!("State requires ErrorEquals") if !@error_equals.kind_of?(Array) || @error_equals.empty?
```

kbrock:
update:
- updated an `error!` to a `missing_field!` (same output, just using the correct method)

lib/floe/validation_mixin.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    private
+
+    def full_name_string
+      full_name.join(".")
+    end
+
+    module ClassMethods
+      def parser_error!(full_name, comment)
+        full_name = full_name.join(".") if full_name.kind_of?(Array)
```
agrare, line 33: Do we need this and also `#full_name_string`?  Could we enforce one or the other or are there some cases where we have to pass an array in?

lib/floe/workflow/state.rb
```diff
@@ -4,32 +4,31 @@ module Floe
   class Workflow
     class State
       include Logging
+      include ValidationMixin
 
       class << self
-        def build!(workflow, name, payload)
+        def build!(workflow, full_name, payload)
```
agrare, line 10: @kbrock do we care about name vs full_name?  Could we keep it as name but have it be the full path instead?

lib/floe/workflow/state.rb
```diff
@@ -4,32 +4,31 @@ module Floe
...
           state_type = payload["Type"]
-          raise Floe::InvalidWorkflowError, "Missing \"Type\" field in state [#{name}]" if payload["Type"].nil?
+          parser_missing_field!(full_name, "Type") if payload["Type"].nil?
```
agrare, line 12: Just naming stuff, but wonder if validate_missing_field! would be better?

kbrock: This is not checking values but rather throwing an error.
It is essentially a wrapper that generates the proper text.

(I would prefer if all/most errors were of the form "$State field $Field $message", but had trouble keeping this language active vs passive, so I punted to `statelint` wording, and kept this as a place holder to fix.)

I will look for a better phrase

lib/floe/validation_mixin.rb
kbrock: Good call. Workflow is a little funky with this, but the code seems fine dropping this extra concept

lib/floe/workflow/state.rb
kbrock: There are a few spots that reference the state's name and want the shorter name.
I've introduced short_name, so I guess we still needed this concept.

So we now have short_name, name (full name), and long_name (type: name)
Maybe we want to drop long_name at a later time

lib/floe/workflow/state.rb
kbrock: going to change to `missing_field_error!`

kbrock:
update:
- renamed full_name to name (added short_name for those that required just the current name)
- updated error names to reflect they are throwing an error
- dropped full_name_str (it was handled by a single join)
- `wrap_parser_error` added localized rescues for parsing paths and the like - will circle back for this.
- default Workflow name to `"State Machine"` (so we could drop `full_name_str`)
- move some workflow specs to the state they are testing
- Add tests around Next, End, missing Task#Resource

update:
- cops

agrare (approved):

PR #228, opened 2024-06-21
[WIP] Workflow output should be JSON

agrare:
Alternate title, JSON all the things

Built on:
* https://github.com/ManageIQ/floe/pull/225
* https://github.com/ManageIQ/floe/pull/227

PR #230, opened 2024-06-26
Output should be in json

kbrock:
## Overview

Extracts/Overlaps with:
- #228
- #222

Changes tests to ensure our `$stdout` and logs are producing json and not ruby.

## Tangent

What we do know:

- Input should be in JSON.
- Output should be in JSON.
- Logs should use JSON.
- Task input and output should use JSON.
- We need better error messaging around bad `input` and `context`, specifically handling `JSON::ParserError`.

What we don't know:

- Who converts `input` from JSON to ruby? (`CLI`, `Context#initialize`, `State`)
- Should the `Context#input` hold JSON or ruby?
- Should the `Context#output` hold JSON or ruby?
- Should `Context` in the database be JSON or ruby?

### DIFF
--- a/lib/floe/cli.rb
+++ b/lib/floe/cli.rb
@@ -30,7 +30,7 @@ module Floe
       # Display status
       workflows.each do |workflow|
         puts "", "#{workflow.name}#{" (#{workflow.status})" unless workflow.context.success?}", "===" if workflows.size > 1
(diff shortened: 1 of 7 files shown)


agrare (approved):
This does seem like a good middle ground, unless we find in the future that state input has to be JSON this should be sufficient

PR #233, opened 2024-07-01
Example intergration tests

kbrock:
fixtures / integration tests

## Overview

A number of asl files are working differently in floe from AWS State Language runtime.

Here are some fixtures to ensure that both environments work the same.

## File Naming

- `test-[state]-[status]-comment`

#### State

The `State` is the `State` under test.
We currently have tests for `Pass`, `Fail`, `Task`, and `Choice`.

#### Status

- The `pass` files contain a number of tests run inline.
- The `fail` tests take a parameter like `{"test":1}` to subequent tests. Since these throw an uncatchable runtime error, we need to execute each individually.
- The `invalid` tests have a parser error. They are not json errors, but rather semantic errors.

#### runner script

I currently have a runner script [ref]. Though working, it is still a WIP. I can add to this repo if desired.

#### Tasks

I have tests for tasks, but am unsure the best way to post the hello world container. It outputs to stderr, stdout, and changes the return code.

Can't remember where we had agreed upon putting this.

[ref]: https://gist.github.com/kbrock/bd4461d4316375a433976043e6472d09

### DIFF
--- a/spec/fixtures/test-choice-fail.asl
+++ b/spec/fixtures/test-choice-fail.asl
@@ -1,0 +1,103 @@
+{
+  "Comment": "Testing Choice failures - sample input: {test: 1}",
+  "StartAt": "Setup",
+  "States": {
+    "Setup": {
+      "Comment": "(Implied test) Result => output:Result",
+      "Type": "Pass",
+      "Next": "ChooseTest",
+      "Parameters": {"test.$": "$.test", "num1": 1, "str1": "1", "num2": 1, "str2": "1"}
+    },
+    "ChooseTest": {
+      "Type":  "Choice",
+      "Choices": [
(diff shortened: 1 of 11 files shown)


Fryguy:
If these are meant for tests, can we move them under the spec directory?  I think of examples as more user-facing references.  I don't think users _want_ to create invalid asls, which is what most of these are.

kbrock:
These are use cases to testing that we are produce the same errors / results as aws state machines.

I'll move to tests

kbrock:
update:
- moved new `examples` into `spec/fixtures` (the existing ones in there are actual examples)

kbrock:
I'm keeping a running tally of examples to validate our behavior in the future

PR #236, opened 2024-07-02
Evaluate IntrinsicFunctions from PayloadTemplate

agrare:

### DIFF
--- a/lib/floe/workflow/intrinsic_function.rb
+++ b/lib/floe/workflow/intrinsic_function.rb
@@ -2,13 +2,28 @@ require "parslet"
 
 module Floe
   class Workflow
-    module IntrinsicFunction
-      def self.evaluate(payload, context = {}, input = {})
+    class IntrinsicFunction
+      def self.value(payload, context = {}, input = {})
+        new(payload).value(context, input)
+      end
+
+      def self.intrinsic_function?(payload)
+        payload.start_with?("States.")
+      end
+
+      attr_reader :payload
+
+      def initialize(payload)
+        @payload = payload
+      end
+
+      def value(context = {}, input = {})
         begin
           tree = Parser.new.parse(payload)
         rescue Parslet::ParseFailed => err
           raise Floe::InvalidWorkflowError, err.message
         end
+
         Floe.logger.debug { "Parsed intrinsic function: #{payload.inspect} => #{tree.inspect}" }
 
         Transformer.new.apply(tree, :context => context, :input => input)
--- a/lib/floe/workflow/payload_template.rb
+++ b/lib/floe/workflow/payload_template.rb
@@ -42,7 +42,7 @@ module Floe
       end
 
       def parse_payload_string(value)
-        return Path.new(value)              if value.start_with?("$")
+        return Path.new(value)              if Path.path?(value)
         return IntrinsicFunction.new(value) if IntrinsicFunction.intrinsic_function?(value)
 
         value
--- a/lib/floe/workflow/payload_template.rb
+++ b/lib/floe/workflow/payload_template.rb
@@ -42,14 +42,17 @@ module Floe
       end
 
       def parse_payload_string(value)
-        value.start_with?("$") ? Path.new(value) : value
+        return Path.new(value)              if value.start_with?("$")
+        return IntrinsicFunction.new(value) if IntrinsicFunction.intrinsic_function?(value)
+
+        value
       end
 
       def interpolate_value(value, context, inputs)
         case value
-        when Array then interpolate_value_array(value, context, inputs)
-        when Hash  then interpolate_value_hash(value, context, inputs)
-        when Path  then value.value(context, inputs)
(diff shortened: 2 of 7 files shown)


lib/floe/workflow/intrinsic_function.rb
```diff
@@ -1,7 +1,24 @@
...
+      def self.value(payload, context = {}, input = {})
+        new(payload).value(context, input)
+      end
+
+      def self.intrinsic_function?(payload)
+        Parser.new.parse(payload)
+        true
+      rescue Parslet::ParseFailed
+        false
+      end
```
agrare, line 13: @Fryguy not sure if you had a better idea for identifying an IntrinsicFunction

lib/floe/workflow/intrinsic_function.rb
```diff
@@ -1,7 +1,24 @@
...
+
+      attr_reader :payload
+
+      def initialize(payload)
+        @payload = payload
+      end
+
+      def value(context = {}, input = {})
```
agrare, line 21: NOTE this is more in line how we handle Path types

lib/floe/workflow/payload_template.rb
```diff
@@ -42,14 +42,17 @@ def parse_payload_hash(value)
       end
 
       def parse_payload_string(value)
-        value.start_with?("$") ? Path.new(value) : value
+        return Path.new(value)              if value.start_with?("$")
```
agrare, line 45: TODO we could move this logic into Path as well like we do below

lib/floe/workflow/intrinsic_function.rb
Fryguy: Yeah, for now I think it's simpler to do `payload.starts_with?("States.")`

Fryguy:
```suggestion
      def self.intrinsic_function?(payload)
        payload.start_with?("States.")
      end
```

Fryguy: Can you also add a spec for this method to the intrinsic_function_spec.rb?

lib/floe/workflow/intrinsic_function.rb
Fryguy: Yeah I noticed that when I started using Path, so this is much more consistent 👍

spec/workflow/payload_template_spec.rb
```diff
@@ -20,6 +20,16 @@
       end
     end
 
+    context "with intrinsic functions" do
+      let(:payload) { {"uuid.$" => "States.UUID()"} }
+      let(:context) { {} }
+      let(:inputs)  { {} }
+
+      it "returns the value from the inputs" do
```
Fryguy, line 28: this statement is slightly off - I think a better test for "value from the inputs" is to use maybe `States.Array($.input)` and give it actual input.  Otherwise this is more like "executes the intrinsic function".

spec/workflow/payload_template_spec.rb
```diff
@@ -20,6 +20,16 @@
...
+        expect(subject.value(context, inputs)["uuid"]).to match(/^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/)
```
Fryguy, line 29: Super-minor, but I have a slightly smaller regex that I used in the uuid specs.  For consistency let's reuse that one (or we can move into a spec/support thing or perhaps use rspec-uuid gem that includes a matcher).  Not necessary for this PR though.

Fryguy:
Overall, LGTM - minor comments above aside from simpler the predicate method contents.

spec/workflow/payload_template_spec.rb
agrare: :+1: done

spec/workflow/payload_template_spec.rb
agrare: Oh yeah my bad copy & paste from below, fixed

lib/floe/workflow/payload_template.rb
agrare: Done

Fryguy (approved):

Fryguy:
Gah I think this conflicted with the merge of #235

lib/floe/workflow/intrinsic_function.rb
```diff
@@ -2,13 +2,28 @@
...
         begin
           tree = Parser.new.parse(payload)
         rescue Parslet::ParseFailed => err
           raise Floe::InvalidWorkflowError, err.message
         end
```
agrare, line 25: @Fryguy maybe in a follow-up but what do you think about doing the Parser.new.parse in the initialize?  It looks like it would always be the same for every invocation of value plus we like raising invalid workflow exception when building the Workflow instance not at runtime as much as possible

PR #238, opened 2024-07-02
Normalize functions to all take arguments

Fryguy:
With all functions able to take parameters, then we can do argument checking in the transformer. This commit adds argument checking for States.UUID.

@agrare Please review.

I did this PR separate as a precursor to States.ArrayPartition where I actually have to check arity and specific argument types.  States.UUID shows how that can be achieved.

Not sure if we want to raise ArgumentError or something else? Thoughts?


agrare:
I think passing arguments to a function that doesn't take one is a payload problem not a runtime problem (like an argument being a string instead of an int or something) so InvalidWorkfloeError would make sense

Fryguy:
@agrare Maybe?  If it's hardcoded I would agree, but if it's inferred, it's a runtime problem I think.

Compare `States.ArrayContains(States.Array(1, 2, 3), -1)` vs `States.ArrayContains($.input, $.target)` with an input of `{:input => [1, 2, 3], :target => -1}`.  The latter is a runtime error I would think not a workflow error.

kbrock:
I've found it is hard to guess when aws will throw a runtime exception and when they will not. In some cases, doing a StringEquals doesn't throw an exception when a StringGreaterThan will. (where the arguments are a string and an integer) -- you can see a few examples in that fixtures file I had.

I think using guard clauses (via Choice) and checking types is the only way around these nuances.

Fryguy:
I wasnt trying to follow aws here, but maybe that's a good idea to see what the simulator does

Fryguy:
@agrare if it's ok, can we merge for now with the regular ArgumentError?  Reason is I'm starting to do others and patterns are emerging. I think it's clearer what choices to make with more examples.  I'm also seeing that I can normalize most of the arity and type checks.

agrare:
> Compare States.ArrayContains(States.Array(1, 2, 3), -1) vs States.ArrayContains($.input, $.target) with an input of {:input => [1, 2, 3], :target => -1}. The latter is a runtime error I would think not a workflow error.

I was thinking more along the lines of if someone wrote `States.UUID(1)` that for sure is a definition error not a runtime error

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -30,9 +30,16 @@ def self.resolve_args(args)
...
         rule(:jsonpath => simple(:v)) { Floe::Workflow::Path.value(v.to_s, context, input) }
 
-        rule(:states_array => {:args => subtree(:args)}) { Transformer.resolve_args(args) }
+        rule(:states_array => {:args => subtree(:args)}) do
+          Transformer.resolve_args(args)
+        end
+
+        rule(:states_uuid => {:args => subtree(:args)}) do
+          args = Transformer.resolve_args(args())
+          raise ArgumentError, "wrong number of arguments to States.UUID (given #{args.size}, expected 0)" unless args.empty?
```
agrare, line 39: I do think this should be `Floe::InvalidWorkflowError` instead of `ArgumentError`, or maybe we catch `ArgumentError` while evaluating the intrinsic_function and raise InvalidWorkflowError then?

agrare (approved):
I think we should tweak what exceptions we're raising so we aren't exposing "ruby" internals to the end user but can do that in a follow-up

lib/floe/workflow/intrinsic_function/transformer.rb
agrare: > I check arity (which I agree is a workflow error), then types (which feels more like a runtime error), and then specific conditions like argument must be non-zero positive (which also feels like a runtime error)

Completely agree on these

PR #242, opened 2024-07-03
Add intrinsic functions

Fryguy:
- States.ArrayPartition
- States.ArrayContains
- States.ArrayRange
- States.ArrayGetItem
- States.ArrayLength
- States.ArrayUnique
- States.Base64Encode
- States.Base64Decode
- States.Hash
- States.MathRandom
- States.MathAdd
- States.StringSplit

Part of #64

### DIFF
--- a/lib/floe/workflow/intrinsic_function/transformer.rb
+++ b/lib/floe/workflow/intrinsic_function/transformer.rb
@@ -153,6 +153,12 @@ module Floe
           Random.new(seed).rand(range_start..range_end)
         end
 
+        rule(:states_math_add => {:args => subtree(:args)}) do
+          args = Transformer.process_args(args(), "States.MathAdd", [Integer, Integer])
+
+          args.sum
+        end
+
         rule(:states_uuid => {:args => subtree(:args)}) do
           Transformer.process_args(args, "States.UUID", [])
 
--- a/lib/floe/workflow/intrinsic_function/transformer.rb
+++ b/lib/floe/workflow/intrinsic_function/transformer.rb
@@ -34,6 +34,18 @@ module Floe
           Transformer.resolve_args(args)
         end
 
+        rule(:states_array_partition => {:args => subtree(:args)}) do
+          args = Transformer.resolve_args(args())
+          raise ArgumentError, "wrong number of arguments to States.ArrayPartition (given #{args.size}, expected 2)" unless args.size == 2
+
+          array, chunk = *args
+          raise ArgumentError, "wrong type for first argument to States.ArrayPartition (given #{array.class}, expected Array)" unless array.kind_of?(Array)
+          raise ArgumentError, "wrong type for second argument to States.ArrayPartition (given #{chunk.class}, expected Integer)" unless chunk.kind_of?(Integer)
+          raise ArgumentError, "invalid value for second argument to States.ArrayPartition (given #{chunk}, expected a positive Integer)" unless chunk.positive?
+
+          array.each_slice(chunk).to_a
+        end
+
         rule(:states_uuid => {:args => subtree(:args)}) do
           args = Transformer.resolve_args(args())
           raise ArgumentError, "wrong number of arguments to States.UUID (given #{args.size}, expected 0)" unless args.empty?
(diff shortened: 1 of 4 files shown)


Fryguy:
@agrare I'd like to merge this one at this point.  It's not all of the functions, but the next few are the more complicated ones.

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -3,41 +3,189 @@
...
+            Base64.strict_decode64(str)
+          rescue ArgumentError => err
+            raise ArgumentError, "invalid value for argument 1 to States.Base64Decode (#{err})"
+          end
+        end
+
+        rule(:states_hash => {:args => subtree(:args)}) do
+          args = Transformer.process_args(args(), "States.Hash", [Object, String])
+          data, algorithm = *args
+          raise NotImplementedError if data.kind_of?(Hash)
```
Fryguy, line 138: Just calling this out - I can't figure out how the stepfunctions simulator converts this to get a Hash but it does, so I just marked as Not implemented for now.

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -3,41 +3,189 @@
...
+            end
           end
         end
 
         rule(:null_literal  => simple(:v)) { nil }
         rule(:true_literal  => simple(:v)) { true }
         rule(:false_literal => simple(:v)) { false }
 
-        rule(:string   => simple(:v)) { v.to_s }
+        rule(:string   => simple(:v)) { v.to_s[1..-2] }
```
agrare, line 62: I'm assuming this is to extract `foo` from `"foo"` ?

lib/floe/workflow/intrinsic_function/transformer.rb
agrare: I wonder if they .to_json the hash then checksum that string?

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -3,41 +3,189 @@
...
+            raise ArgumentError, "invalid value for argument 1 to States.Hash (given null, expected non-null)"
+          end
+
+          algorithms = %w[MD5 SHA-1 SHA-256 SHA-384 SHA-512]
+          unless algorithms.include?(algorithm)
+            raise ArgumentError, "invalid value for argument 2 to States.Hash (given #{algorithm.inspect}, expected one of #{algorithms.map(&:inspect).join(", ")})"
+          end
+
+          require "openssl"
+          algorithm = algorithm.sub("-", "")
```
agrare, line 149: Very minor but we could have a static hash of their algo names mapped to `OpenSSL::Digest` algo names, could use .keys for the check above and avoid an array iteration here

lib/floe/workflow/intrinsic_function/transformer.rb
Fryguy: Yes, so I had to make a change because empty string wasn't being handled properly.  More details in the commit message here: https://github.com/ManageIQ/floe/pull/242/commits/3a3640650f8521e899d7ff3184b7554d69aacc8d

>  Fix issue where empty string would not parse as expected
>
> Normal string would parse into a slice, however an empty string would
> parse into an empty array. Thus the transformer would not match because
> simple can only match on scalar values. Instead, we now include the
> quotes in the string and remove them afterwards.

lib/floe/workflow/intrinsic_function/transformer.rb
Fryguy: Oh hadn't though of checksum - I tried json, json without spaces, pretty json

agrare (approved):
LGTM nothing significant blocking this

lib/floe/workflow/intrinsic_function/transformer.rb
Fryguy: Yeah I can do that.  It turns out I only need this for Ruby 2.7 support - In Ruby 3.0 the full name just works.  Would you prefer the hash lookup, or maybe I can just gate this with a Ruby < 3.0 check?

lib/floe/workflow/intrinsic_function/transformer.rb
agrare: Hm yeah no variation of JSON of the hash that I tried seems to match what you have in the spec (I assume that value came from the step-functions simulator)

PR #243, opened 2024-07-08
Fix sporadic test failure with Wait state

agrare:
I suspect there is an issue if the GHA runner is a little slow that the `workflow.run_nonblock` can finish before the `expect(workflow.end?).to eq(false)` can be invoked within 1 second.

https://github.com/ManageIQ/floe/actions/runs/9845986602/job/27182755884?pr=184
```
Failures:
  1) Floe::Workflow::States::Pass#running? with timestamp runs
     Failure/Error: expect(workflow.end?).to eq(false)

       expected: false
            got: true

Failed examples:

rspec ./spec/workflow/states/wait_spec.rb:94 # Floe::Workflow::States::Pass#running? with timestamp runs
```

### DIFF
--- a/spec/workflow/states/wait_spec.rb
+++ b/spec/workflow/states/wait_spec.rb
@@ -2,7 +2,7 @@ RSpec.describe Floe::Workflow::States::Pass do
   let(:input)    { {} }
   let(:ctx)      { Floe::Workflow::Context.new(:input => input.to_json) }
   let(:state)    { workflow.start_workflow.current_state }
-  let(:workflow) { make_workflow(ctx, {"WaitState" => {"Type" => "Wait", "Seconds" => 1, "Next" => "SuccessState"}, "SuccessState" => {"Type" => "Succeed"}}) }
+  let(:workflow) { make_workflow(ctx, {"WaitState" => {"Type" => "Wait", "Seconds" => 10, "Next" => "SuccessState"}, "SuccessState" => {"Type" => "Succeed"}}) }
 
   describe "#end?" do
     it "is non-terminal" do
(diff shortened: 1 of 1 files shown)


spec/workflow/states/wait_spec.rb
```diff
@@ -73,10 +99,23 @@
...
+        Timecop.travel(Time.now.utc + 100) do
+          workflow.run_nonblock
+        end
+
+        expect(workflow.end?).to eq(true)
+        expect(workflow.context.state_history.size).to eq(2)
+      end
     end
 
-    context "with timestamp" do
```
agrare, line 78: NOTE this context was confusing because it is the same as above, believe it should have been TimestampPath

spec/workflow/states/wait_spec.rb
```diff
@@ -73,10 +99,23 @@
...
+        expect(workflow.end?).to eq(false)
+        expect(workflow.context.state_history.size).to eq(0)
+
+        Timecop.travel(Time.now.utc + 100) do
+          workflow.run_nonblock
+        end
+
+        expect(workflow.end?).to eq(true)
+        expect(workflow.context.state_history.size).to eq(2)
+      end
```
agrare, line 114: These other contexts didn't have the `it "runs"` like TimestampPath did.

Fryguy: total aside, you can also just do `it do` without any text.

kbrock: I've started really liking `it do` - especially when we have it runs / it throws an exception

agrare: Or I'll just give it a more descriptive name

PR #245, opened 2024-07-09
Move the Parslet parse into initialize so that invalid function definition will fail on workflow load

agrare:
I still need to handle the exception types for other failures, but this is a quick way to raise an exception earlier than runtime for a number of failure types

### DIFF
--- a/lib/floe/workflow/intrinsic_function.rb
+++ b/lib/floe/workflow/intrinsic_function.rb
@@ -15,19 +15,20 @@ module Floe
 
       def initialize(payload)
         @payload = payload
+        @tree    = Parser.new.parse(payload)
+
+        Floe.logger.debug { "Parsed intrinsic function: #{payload.inspect} => #{tree.inspect}" }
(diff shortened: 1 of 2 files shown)


lib/floe/workflow/intrinsic_function.rb
```diff
@@ -15,19 +15,20 @@ def self.intrinsic_function?(payload)
...
       end
 
       def value(context = {}, input = {})
-        begin
-          tree = Parser.new.parse(payload)
-        rescue Parslet::ParseFailed => err
-          raise Floe::InvalidWorkflowError, err.message
-        end
-
         Floe.logger.debug { "Parsed intrinsic function: #{payload.inspect} => #{tree.inspect}" }
```
Fryguy, line 24: Should we also move this debug logging to the initialize method?  It's meant to show the parsed tree, so it belongs with parsing.

agrare: :+1: done

Fryguy (approved):

PR #248, opened 2024-07-14
Add spec/supports

kbrock:
I tend to want to require "byebug"
I put this into the supports directory (and don't commit)

This allows this and other local enhancement behavior

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -59,12 +59,7 @@ RSpec.configure do |config|
   config.include AwesomeSpawn::SpecHelper, :uses_awesome_spawn => true
   config.before(:each, :uses_awesome_spawn) { disable_spawning }
 
-  # factory methods
-
-  def make_workflow(ctx, states)
-    payload = {"StartAt" => states.keys.first, "States" => states}
-    Floe::Workflow.new(payload, ctx)
-  end
+  config.include CommonMethods
 end
 
 require "floe"
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -18,6 +18,8 @@ if ENV['CI']
   SimpleCov.start
 end
 
+Dir['./spec/support/**/*.rb'].sort.each { |f| require f }
+
 # See https://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 RSpec.configure do |config|
   # rspec-expectations config goes here. You can use an alternate
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -61,18 +61,9 @@ RSpec.configure do |config|
 
   # factory methods
 
-  def make_workflow(ctx, payload)
-    Floe::Workflow.new(make_payload(payload), ctx)
-  end
-
-  def make_payload(states)
-    start_at ||= states.keys.first
-
-    {
-      "Comment" => "Sample",
-      "StartAt" => start_at,
-      "States"  => states
-    }
+  def make_workflow(ctx, states)
+    payload = {"StartAt" => states.keys.first, "States" => states}
+    Floe::Workflow.new(payload, ctx)
   end
 end
 
--- a/spec/support/common_methods.rb
+++ b/spec/support/common_methods.rb
@@ -1,0 +1,8 @@
+module CommonMethods
+  # factory methods
+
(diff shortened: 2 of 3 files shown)


Fryguy:
We do need spec/supports for other stuff though - think you had another PR where you wanted to introduce example files, and a helper method around it.  Also there are [helper methods directly in the spec_helper.rb](https://github.com/ManageIQ/floe/blob/master/spec/spec_helper.rb#L60-L74) that should probably be moved to a proper spec/supports file.

kbrock:
>  I end up having to `require "byebug"; byebug`

```
$ cat ~/.bundler.d/Gemfile.global.rb
ensure_gem "byebug"
```

Yea, that is my exact use case. I have added a `supports/byebug.rb` file to all my other projects:

```bash
echo "require 'byebug'" > spec/supports/byebug.rb
```

So in code I can just type `byebug`

kbrock:
update:
- moved factory methods into `spec/supports`

spec/spec_helper.rb
```diff
@@ -57,21 +59,7 @@
...
-  def make_payload(states)
-    start_at ||= states.keys.first
-
-    {
-      "Comment" => "Sample",
-      "StartAt" => start_at,
-      "States"  => states
-    }
-  end
+  config.include FloeFactory
```
agrare, line 62: Just naming, but what about `FloeHelper` instead?  Especially in specs Factory makes me think of FactoryBot which this isn't and AwesomeSpawn have `AwesomeSpawn::SpecHelper` above

Fryguy: Or `Floe::SpecHelper`?

agrare: Yeah wasn't sure if he was trying to keep it split out into specific concerns, maybe `Floe::SpecHelper` if you want all in one or `Floe::WorkflowHelper` if you want to keep things split out

kbrock: `AwesomeSpawn::SpecHelper` is used by non-AwesomeSpawn code to help better leverage the `AwesomeSpawn` library.

So if we provided helpers to be used by the workflow or other non-floe code, then that could be a great option.

So I'm still not sure what to call this.

kbrock:
> [Line 21](https://github.com/kbrock/floe/blob/cd7071acc6a45fb9cfcfdc6327d66ace17bb6db9/spec/spec_helper.rb#L21), Col 1 - [Lint/NonDeterministicRequireOrder](https://docs.rubocop.org/rubocop/1.56/cops_lint.html#lintnondeterministicrequireorder) - Sort files before requiring them.

omg. I just changed this line because it said it was bad to sort first.
The conflicting warnings are frustrating

spec/spec_helper.rb
Fryguy: @kbrock That's a good point we should reserve Floe::SpecHelper for a public module we might want to share for callers to test floe stuff in the same way we have AwesomeSpawn::SpecHelper

I checked our other gems, and everywhere else, we either just call it by feature name followed by helper (not scoped) (e.g. `DatabaseHelper`, `ConnectionHelper`), or we have it in a Spec::Support namespect (e.g. `Spec::Support::DatabaseHelper`). Helper as a term is pretty consistent.

So I'm thinking spec/support/workflow_helper.rb and inside either WorkflowHelper or Spec::Support::WorkflowHelper.  Honestly though, I'm fine with anything

Fryguy:
> it said it was bad to sort first.

I don't know what the code was doing before, but I assume you were hitting Lint/RedundantDirGlobSort ?

Lint/NonDeterministicRequireOrder calls that out kind of

> Dir.glob and Dir[] sort globbed results by default in Ruby 3.0. So all bad cases are acceptable when Ruby 3.0 or higher are used.
> Note: This cop will be deprecated and removed when supporting only Ruby 3.0 and higher.

So I think we can ignore (or mark it as disabled if we really care).

spec/spec_helper.rb
kbrock: I tried out `CommonMethods` but I do like the `Helpeb word suggestion.
I thought we didn't do namespacing.

kbrock:
update:
- fixed typo: "Type":"Succees" => "Type":"Succeed"

agrare (approved):

PR #251, opened 2024-07-18
Add shared example to cut down on Wait spec duplication

kbrock:
Not sure if it is more readable - or if we even consider the previous code having duplication


kbrock:
update:
- rebased

Fryguy (approved):

PR #253, opened 2024-08-01
Choice rule payload validation path

kbrock:
extracted from #189 

# Before

If the path referenced a missing value in the input, treat it as a nil.

# After

If the path references a missing value in the input, raise an error.

- introduced `presence_check` method to check if it is present up front.
- This also affects Intrinsic methods. You can see the changed spec.
- This respects `IsPresent: false` and `IsPresent: true`

### DIFF
--- a/spec/workflow/path_spec.rb
+++ b/spec/workflow/path_spec.rb
@@ -11,15 +11,15 @@ RSpec.describe Floe::Workflow::Path do
   describe "#value" do
     context "referencing the global context" do
       it "with a missing value" do
-        expect(described_class.new("$$.foo").value({})).to be_nil
+        expect { described_class.new("$$.foo").value({}, {"foo" => "bar"}) }.to raise_error(Floe::PathError, "Path [$$.foo] references an invalid value")
       end
 
       it "with a single value" do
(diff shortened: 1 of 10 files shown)


spec/workflow/path_spec.rb
```diff
@@ -11,15 +11,15 @@
   describe "#value" do
     context "referencing the global context" do
       it "with a missing value" do
-        expect(described_class.new("$$.foo").value({})).to be_nil
+        expect { described_class.new("$$.foo").value({}, {"foo" => "bar"}) }.to raise_error(Floe::PathError, "Path [$$.foo] references an invalid value")
```
kbrock, line 14: this is the essence of the issue

kbrock:
update:
- removed spec updates/refactor (I'll put in another PR)

kbrock:
update:
- removed changes to `is_present?` -- it is not useful now, will delete after the dust settles.

kbrock:
WIP: moving stuff around. Want to get other stuff merged before this.
There is no outstanding work for this item.

kbrock:
WIP: oops. thought this was something else.
This is good to go

agrare (approved):
:+1: LGTM

PR #255, opened 2024-08-08
Intrinsics States.JsonMerge

kbrock:
Part of #64

Ruby missing a `Boolean` is frustrating me more and more.

This adds `JsonMerge`.

Please share your thoughts on how to handle the Boolean Attribute.

The other solution I had was a multi argument that handles a crystal style argument and it went through each of them. But it still had more modification in the comparison portion of the check code

Another solution is to just specify Object, and then manually make a check, but this required duplication of the error message.

### DIFF
--- a/lib/floe/workflow/intrinsic_function/parser.rb
+++ b/lib/floe/workflow/intrinsic_function/parser.rb
@@ -67,6 +67,7 @@ module Floe
           :states_base64_encode,   "States.Base64Encode",
           :states_base64_decode,   "States.Base64Decode",
           :states_hash,            "States.Hash",
+          :states_json_merge,      "States.JsonMerge",
           :states_math_random,     "States.MathRandom",
           :states_math_add,        "States.MathAdd",
           :states_string_split,    "States.StringSplit",
@@ -93,6 +94,7 @@ module Floe
             states_base64_encode |
             states_base64_decode |
             states_hash |
+            states_json_merge |
             states_math_random |
             states_math_add |
             states_string_split |
--- a/lib/floe/workflow/intrinsic_function/transformer.rb
+++ b/lib/floe/workflow/intrinsic_function/transformer.rb
@@ -193,6 +193,18 @@ module Floe
           OpenSSL::Digest.hexdigest(algorithm, data)
         end
 
+        rule(:states_json_merge => {:args => subtree(:args)}) do
+          args = Transformer.process_args(args(), "States.JsonMerge", [Hash, Hash, [TrueClass, FalseClass]])
+          left, right, deep = *args
+
+          if deep
+            # NOTE: not implemented by aws States language and nuances not defined in docs
+            left.merge(right) { |_key, l, r| l.kind_of?(Hash) && r.kind_of?(Hash) ? l.merge(r) : r }
+          else
+            left.merge(right)
+          end
+        end
+
         rule(:states_json_to_string => {:args => subtree(:args)}) do
           args = Transformer.process_args(args(), "States.JsonToString", [Object])
           json = args.first
--- a/spec/workflow/intrinsic_function_spec.rb
+++ b/spec/workflow/intrinsic_function_spec.rb
@@ -529,6 +529,35 @@ RSpec.describe Floe::Workflow::IntrinsicFunction do
       end
     end
 
+    describe "States.JsonMerge" do
+      it "it merges with right hand precedence" do
+        result = described_class.value(
+          "States.JsonMerge($.left, $.right, false)", {},
+          {"left" => {"a" => "la", "b" => "lb"}, "right" => {"b" => "rb"}}
+        )
+        expect(result).to eq({"a" => "la", "b" => "rb"})
+      end
+
+      it "it deep merges with right hand precedence" do
+        result = described_class.value(
+          "States.JsonMerge($.left, $.right, true)", {},
+          {"left" => {"b" => {"ba" => "lb", "bb" => "lb"}}, "right" => {"b" => {"ba" => "rb"}}}
+        )
+        expect(result).to eq({"b" => {"ba" => "rb", "bb" => "lb"}})
+      end
+
+      it "fails with wrong number of parameters" do
+        expect { described_class.value("States.JsonMerge($.left, $.right)", {}, {"left" => [1, 2], "right" => {"a" => "la"}}) }.to raise_error(ArgumentError, "wrong number of arguments to States.JsonMerge (given 2, expected 3)")
+        expect { described_class.value("States.JsonMerge($.left, $.right, false, 5)", {}, {"left" => [1, 2], "right" => {"a" => "la"}}) }.to raise_error(ArgumentError, "wrong number of arguments to States.JsonMerge (given 4, expected 3)")
+      end
+
+      it "fails with wrong type of parameters" do
+        expect { described_class.value("States.JsonMerge($.left, $.right, false)", {}, {"left" => [1, 2], "right" => {"a" => "la"}}) }.to raise_error(ArgumentError, "wrong type for argument 1 to States.JsonMerge (given Array, expected Hash)")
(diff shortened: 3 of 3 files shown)


lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -9,6 +9,11 @@ class Workflow
     class IntrinsicFunction
       class Transformer < Parslet::Transform
         OptionalArg = Struct.new(:type)
+        class Boolean
```
Fryguy, line 12: Super-mega-minor
```suggestion

        class Boolean
```

Fryguy:
Wow...it's that simple!?  LOL I was so scared to do this one haha.

Fryguy:
oops...accidental close.

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -9,6 +9,11 @@ class Workflow
...
+          def self.to_s ; "Boolean" ; end
+          # NOTE: this is backwards
+          def self.kind_of?(value) ; value.kind_of?(TrueClass) || value.kind_of?(FalseClass) ; end
```
Fryguy, line 15: I agree with Rubocop - not a fan of single-line methods, even if simple, so let's just blow this out.

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -9,6 +9,11 @@ class Workflow
     class IntrinsicFunction
       class Transformer < Parslet::Transform
         OptionalArg = Struct.new(:type)
+        class Boolean
+          def self.to_s ; "Boolean" ; end
+          # NOTE: this is backwards
```
Fryguy, line 14: Can you add more detail here?  What makes it backwards in your opinion?  Are there tests demostrating this backwardness?

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -50,7 +55,8 @@ def check_types(args, function, signature)
             args.zip(signature).each_with_index do |(arg, type), index|
               type = type.type if type.kind_of?(OptionalArg)
 
-              raise ArgumentError, "wrong type for argument #{index + 1} to #{function} (given #{arg.class}, expected #{type})" unless arg.kind_of?(type)
+              matched = type == Boolean ? type.kind_of?(arg) : arg.kind_of?(type)
```
Fryguy, line 58: I wonder if we can do this better with something like `===`, which can handle kind_of logic?

spec/workflow/intrinsic_function_spec.rb
```diff
@@ -450,6 +450,23 @@
...
+        result = described_class.value(
+          "States.JsonMerge($.left, $.right, false)",
+          ctx,
+          {"left" => {"a" => "la", "b" => "lb"}, "right" => {"b" => "rb"}}
+        )
+        expect(result).to eq({"a" => "la", "b" => "rb"})
+      end
+
+      it "fails with invalid args" do
+        expect { described_class.value("States.JsonMerge($.left, $.right, 5)", ctx, {"left" => {"a" => "la"}, "right" => {"b" => "rb"}}) }.to raise_error(ArgumentError, "wrong type for argument 3 to States.JsonMerge (given Integer, expected Boolean)")
```
Fryguy, line 466: Please include failure tests for every parameter (i.e. non-Hash in positions 1 and 2)

spec/workflow/intrinsic_function_spec.rb
```diff
@@ -450,6 +450,23 @@
       end
     end
 
+    describe "States.JsonMerge" do
+      let(:ctx) { {"context" => {"baz" => "qux"}} }
```
Fryguy, line 454: This context is not needed, so it complicated the test - Can we remove it and just hardcode an empty context like we do for the other tests?

Fryguy:
I think the addition of Boolean is great.  Only thought I had was maybe to implement `===`, then we can unify the check by switching from kind_of? to `===`.  Not sure it will work, but there's so many tests, that we'll know pretty quickly.  Even so, I'm happy to merge as is, and this can be a followup.

Fryguy:
Overall LGTM :+1: - A few comment on specs, then this is good to go.

lib/floe/workflow/intrinsic_function/parser.rb
```diff
@@ -64,6 +64,7 @@ class Parser < Parslet::Parser
           :states_base64_encode,   "States.Base64Encode",
           :states_base64_decode,   "States.Base64Decode",
           :states_hash,            "States.Hash",
+          :states_json_merge,      "States.JsonMerge",
```
Fryguy, line 67: This order is not alphabetical - it currently matches the order as listed in the stateslanguage.net spec.  The same goes for the order of the methods within the transformer.

lib/floe/workflow/intrinsic_function/transformer.rb
kbrock: we do `Boolean.kind_of?(obj)` instead of `obj.kind_of?(Boolean)`

That is why we have a weird.

I have renamed to `cover?` and it reads a little better

lib/floe/workflow/intrinsic_function/transformer.rb
kbrock: ooh. we are not asking if type is a kind of Boolean, we are asking if type == Boolean.
Introducing `===` makes the tests fail

kbrock:
Looks like we need to introduce `deep_merge` -- not sure if we want to bring in a gem or want to implement this. (I was surprised to see that `deep_merge` gem was one of yours. YAY)

The problem with `===` and `kind_of?` is the method is defined on `Object` not `Boolean`. And we will not get called in those instances.

@Fryguy what do you think about defining `Boolean`?
I'm thinking of adding Boolean to other places too.

kbrock:
update:
- Rebase
- Introduce `Boolean` to remove special cases around `kind_of?`
- Implement `deep_merge`
- Add more specs.

update:
- Fixed spacing in hash (rubocop)

Fryguy:
This need a rebase after #256

spec/workflow/intrinsic_function_spec.rb
```diff
@@ -450,6 +450,30 @@
       end
     end
 
+    describe "States.JsonMerge" do
+      it "it merges with right hand precidence" do
```
Fryguy, line 454:
```suggestion
      it "it merges with right hand precedence" do
```

spec/workflow/intrinsic_function_spec.rb
```diff
@@ -450,6 +450,30 @@
...
+        result = described_class.value(
+          "States.JsonMerge($.left, $.right, false)", {},
+          {"left" => {"a" => "la", "b" => "lb"}, "right" => {"b" => "rb"}}
+        )
+        expect(result).to eq({"a" => "la", "b" => "rb"})
+      end
+
+      it "it deep merges with right hand precidence" do
```
Fryguy, line 462:
```suggestion
      it "it deep merges with right hand precedence" do
```

lib/floe/boolean.rb
```diff
@@ -0,0 +1,10 @@
+module Floe
+  module Boolean
```
Fryguy, line 2: Not sure I like this here - Originally you had it under Floe::Workflow::IntrinsicFunction::Boolean, which made more sense to keep it contained there.  Also, not sure I like messing with the internals of TrueClass and FalseClass.

Would like another opinion though - @agrare ?

kbrock:
Update:
- rebase (resolved conflict)
- fixed spelling
- added number of parameters check

Still unsure what to do about Boolean.
Thinking that a MultiType (ala Crystal `TrueClass|FalseClass` / `[TrueClass, FalseClass]`) may be the best route.

Fryguy:
Coincidentally, I added "Union" types to my States.Format PR, because I needed that.  Perhaps you can use that after that one is merged?

kbrock:
Update:
- dropped Boolean and used Union type

Fryguy (approved):

PR #244, opened 2024-07-08
Add tool for using the aws stepfunctions simulator

Fryguy:
@kbrock @agrare Please take a look. I mostly just use this for testing .asl files and had a special helper for intrinsic functions, but we can expand it as needed in future PRs.

```bash
$ tools/step_functions
Run the aws stepfunctions simulator.

Notes:
  This tool requires the stepfunctions simulator to be installed locally and running.
  Installation instructions can be found at https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local.html.

Commands:
  execute       Execute an .asl file through the stepfunctions simulator.
  intrinsic     Execute an intrinsic function or JSONPath standalone.

  For more help with a specific command use tools/step_functions <command> --help

Global Options:
  -h, --help    Show this message

```

```
$ tools/step_functions execute --help
Execute an .asl file through the stepfunctions simulator.

  -f, --file=<s>     The .asl file to execute (default: definition.asl)
  -i, --input=<s>    Input to the execution
  -h, --help         Show this message
```

```
$  tools/step_functions intrinsic --help
Execute an intrinsic function or JSONPath standalone.

  -f, --function=<s>    The intrinsic function or JSONPath to run
  -i, --input=<s>       Input to the execution
  -h, --help            Show this message
```

PR #256, opened 2024-08-08
Intrinsics JsonToString and StringToJson

kbrock:
part of #64

Not sure if we want to do a special json error parse case

### DIFF
--- a/lib/floe/workflow/intrinsic_function/parser.rb
+++ b/lib/floe/workflow/intrinsic_function/parser.rb
@@ -54,6 +54,8 @@ module Floe
         end
 
         [
+          :states_string_to_json,  "States.StringToJson",
+          :states_json_to_string,  "States.JsonToString",
           :states_array,           "States.Array",
           :states_array_partition, "States.ArrayPartition",
           :states_array_contains,  "States.ArrayContains",
@@ -77,7 +79,9 @@ module Floe
         end
 
         rule(:expression) do
-          states_array |
+          states_string_to_json |
+            states_json_to_string |
+            states_array |
             states_array_partition |
             states_array_contains |
             states_array_range |
--- a/lib/floe/workflow/intrinsic_function/transformer.rb
+++ b/lib/floe/workflow/intrinsic_function/transformer.rb
@@ -151,6 +151,12 @@ module Floe
           OpenSSL::Digest.hexdigest(algorithm, data)
         end
 
+        rule(:states_json_to_string => {:args => subtree(:args)}) do
+          args = Transformer.process_args(args(), "States.JsonToString", [Object])
+          json = args.first
+          JSON.generate(json)
+        end
+
         rule(:states_math_random => {:args => subtree(:args)}) do
           args = Transformer.process_args(args(), "States.MathRandom", [Integer, Integer, OptionalArg[Integer]])
           range_start, range_end, seed = *args
@@ -180,6 +186,14 @@ module Floe
           else
             str.split(/[#{Regexp.escape(delimeter)}]+/)
           end
+        end
+
+        rule(:states_string_to_json => {:args => subtree(:args)}) do
+          args = Transformer.process_args(args(), "States.StringToJson", [String])
+          string = args.first
+          JSON.parse(string)
+        rescue JSON::ParserError => e
+          raise ArgumentError, "invalid json: #{e.message}"
         end
 
         rule(:states_uuid => {:args => subtree(:args)}) do
--- a/spec/workflow/intrinsic_function_spec.rb
+++ b/spec/workflow/intrinsic_function_spec.rb
@@ -450,6 +450,41 @@ RSpec.describe Floe::Workflow::IntrinsicFunction do
       end
     end
 
+    describe "States.JsonToString" do
+      it "fetches values from the input" do
+        # this is not in the spec but automatic as part of the parser
+        result = described_class.value("States.JsonToString(true)", {}, {})
+        expect(result).to eq("true")
+
+        result = described_class.value("States.JsonToString($.input)", {}, {"input" => "foo"})
+        expect(result).to eq("\"foo\"")
+
+        result = described_class.value("States.JsonToString($.input)", {}, {"input" => nil})
+        expect(result).to eq("null")
+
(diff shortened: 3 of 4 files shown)


lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -151,6 +151,11 @@ def check_types(args, function, signature)
           OpenSSL::Digest.hexdigest(algorithm, data)
         end
 
+        rule(:states_json_to_string => {:args => subtree(:args)}) do
+          args = Transformer.process_args(args(), "States.JsonToString", [Object])
+          JSON.generate(args.first)
```
Fryguy, line 156: I know it's pedantic, but for code consistency purposes
```suggestion
          json = args.first

          JSON.generate(json)
```

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -182,6 +187,11 @@ def check_types(args, function, signature)
           end
         end
 
+        rule(:states_string_to_json => {:args => subtree(:args)}) do
+          args = Transformer.process_args(args(), "States.StringToJson", [String])
+          JSON.parse(args.first)
```
Fryguy, line 192: Similarly here:

```suggestion
          string = args.first

          JSON.parse(string)
```

spec/workflow/intrinsic_function_spec.rb
```diff
@@ -450,6 +450,13 @@
       end
     end
 
+    describe "States.JsonToString" do
+      it "fetches values from the input" do
+        result = described_class.value("States.JsonToString($.input)", {"context" => {"baz" => "qux"}}, {"input" => {"foo" => "bar"}})
+        expect(result).to eq("{\"foo\":\"bar\"}")
+      end
```
Fryguy, line 457: Please include a test for the various failure cases (wrong type, too few params, too many params).

spec/workflow/intrinsic_function_spec.rb
```diff
@@ -450,6 +450,13 @@
       end
     end
 
+    describe "States.JsonToString" do
+      it "fetches values from the input" do
+        result = described_class.value("States.JsonToString($.input)", {"context" => {"baz" => "qux"}}, {"input" => {"foo" => "bar"}})
```
Fryguy, line 455: The context is not needed for the test, so it visually compicates the test, so let's get rid of it and replace with a hardcoded `{}` like we do in other specs.

spec/workflow/intrinsic_function_spec.rb
```diff
@@ -610,6 +617,13 @@
       end
     end
 
+    describe "States.StringToJson" do
+      it "fetches values from the input" do
+        result = described_class.value("States.StringToJson($.input)", {"context" => {"baz" => "qux"}}, {"input" => "{\"foo\":\"bar\"}"})
+        expect(result).to eq({"foo" => "bar"})
+      end
```
Fryguy, line 624: Please also add the common failure cases as before.

Addtionally, there can be "bad" strings, so let's have specs for those as well (and on that note, they will probably raise a JSON::ParserError that we probably need to capture and convert to a Floe error.  Offhand, this should non-json strings (e.g. "XXX"), the empty string, a string with only a newline, and non-hash JSON strings like Array, Integer, "null", etc (there's probably more we can verify).

spec/workflow/intrinsic_function_spec.rb
```diff
@@ -610,6 +617,13 @@
       end
     end
 
+    describe "States.StringToJson" do
+      it "fetches values from the input" do
+        result = described_class.value("States.StringToJson($.input)", {"context" => {"baz" => "qux"}}, {"input" => "{\"foo\":\"bar\"}"})
```
Fryguy, line 622: Similarly please drop the context.

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -151,6 +151,11 @@ def check_types(args, function, signature)
           OpenSSL::Digest.hexdigest(algorithm, data)
         end
 
+        rule(:states_json_to_string => {:args => subtree(:args)}) do
+          args = Transformer.process_args(args(), "States.JsonToString", [Object])
```
Fryguy, line 155: If it's Object then we should have a separate test for each type of object possible that can be converted to JSON (might be useful to run this through the stepfunctions simulartor to see what it does - my script in #244 is helpful for this purpose)

lib/floe/workflow/intrinsic_function/transformer.rb
Fryguy, line 156: Is it possible for JSON.generate to fail?  It feels like it could, but I'm not sure how.  If it can fail with a specific error, we will likely need to rescue and wrap that in a Floe error.

lib/floe/workflow/intrinsic_function/parser.rb
```diff
@@ -64,9 +64,11 @@ class Parser < Parslet::Parser
           :states_base64_encode,   "States.Base64Encode",
           :states_base64_decode,   "States.Base64Decode",
           :states_hash,            "States.Hash",
+          :states_json_to_string,  "States.JsonToString",
```
Fryguy, line 67: This order is not alphabetical - it currently matches the order as listed in the stateslanguage.net spec.  The same goes for the order of the methods within the transformer.

Fryguy:
This is also missing parser_spec tests - they are kind of dumb - you can copy another one like States.Array - it's basically just checking a happy path, and then the bad path is when someone puts `StringToJson` without parens.

lib/floe/workflow/intrinsic_function/parser.rb
Fryguy: usually for the happy path test I just use the one example from states-language.net :)

kbrock:
update:
- Rebase.
- Add args parameter names to transformer.
- Add bad argument tests.
- Added all class input for JsonToString.
- Drop unneeded contexts from tests.
- Changed order of transformer to match specifications.

Fryguy (approved):
There a few organizational things that don't match the other specs, but that can all be followups for cleanup, so merging.

Overall, this is great! Thanks @kbrock

PR #257, opened 2024-08-08
In ChoiceRule::Data, parse compare key and variable 

kbrock:
```json
{
  "Choices": [
    {"Variable": "$.foo", "GreaterThanPath": "$.bar", "Next": "Phase2"}
  ]
}
```

## Changes

| Concept | value | change |
|---------|-------|--------|
|`@variable`|`"$.foo"`|Move over to `ChoiceRule::Data`. Store as a `Path`.|
|`@compare_key`|`"GreaterThanPath"`| Validate. Store in variable.|
|`@path`|`true`|Store in variable. `true` when key has `"Path"` in it.|
|`@value`|`"$.bar"`|Store in variable as `String` or `Path` depending upon `@path`.
|`@next`|`"Phase2"`|Validate. Required if top level condition.|

## Changes in words

- Move `Variable` over to `ChoiceRule::Data` since this is only valid for those cases.
- Validates the compare key and keep as a variable.
- Validates the compare key value and keep as a variable.
- If compare key is a path, then `@path = true` and `@value` is a `Path`.
- Validate `Next`.

### DIFF
--- a/lib/floe/workflow/choice_rule.rb
+++ b/lib/floe/workflow/choice_rule.rb
@@ -3,6 +3,8 @@
 module Floe
   class Workflow
     class ChoiceRule
+      include ValidationMixin
+
       class << self
         def build(workflow, name, payload)
           if (sub_payloads = payload["Not"])
@@ -27,11 +29,13 @@ module Floe
 
       attr_reader :next, :payload, :children, :name
 
-      def initialize(_workflow, name, payload, children = nil)
+      def initialize(workflow, name, payload, children = nil)
         @name      = name
         @payload   = payload
         @children  = children
         @next      = payload["Next"]
+
+        validate_next!(workflow)
       end
 
       def true?(*)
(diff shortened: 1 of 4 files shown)


lib/floe/workflow/choice_rule.rb
```diff
@@ -42,8 +44,22 @@ def true?(*)
...
+          missing_field_error!("Next")
+        elsif !workflow_state?(@next, workflow)
+          # top level nodes require a next field that is found
+          invalid_field_error!("Next", @next, "is not found in \"States\"")
+        end
+      end
+
+      # returns true if this is a child rule underneath an And/Or/Not
+      def is_child? # rubocop:disable Naming/PredicateName
+        !(%w[And Or Not] & name[0..-2]).empty?
```
Fryguy, line 62: Is name an Array?  something about the variable "name" is mentally confusing cause it feels like it would be a String, but & can only work with an Array.

Fryguy: Soemthing about this logic is so confusing for me (probably that's why I wrote include_any and include_none in more_core haha).  Can you give an example here what name would look like and how it would match or not match?

spec/workflow/choice_rule_spec.rb
```diff
@@ -16,6 +16,39 @@
     it "works with valid next" do
       workflow
     end
+
+    context "With unknown Next" do
```
Fryguy, line 20: Super-minor
```suggestion
    context "with unknown Next" do
```

lib/floe/workflow/choice_rule.rb
kbrock: I had `%w[And Or Not].include?(name[-3])` but this was confusing to Adam.

```json
{
  "Or": [
    {"Variable": "$.foo", "IsString": true},
    {"Variable": "$.foo", "IsBoolean": true}
  ], "Next": "Finished"
}
```

Basically, the paths will be:
- `[..., "Choices", "Or"]` -- require a "Next".
- `[..., "Choices", "Or", 1, "Data"]` -- no "Next" allowed.
- `[..., "Choices", "Or", 2, "Data"]` -- no "Next" allowed.

Note, the first one has an "Or", but it is not in the parent, hence the `name[0..-2]`

@agrare you still prefer the `& name[0..-2]` over `include?(name[-3])`?

kbrock:
update:
- comments for is_child?
- fixed case of with

kbrock:
update:
- rebased
- moved methods around to reduce size of `diff`. (no actual code changes)

lib/floe/workflow/choice_rule.rb
agrare: Yeah it doesn't read great and my suggestion probably made it worse, I was originally concerned that .include? would be doing a substring search and other names could accidentally match but actually that might not be possible if name is an array and not a flat string.

agrare (approved):

lib/floe/workflow/choice_rule.rb
agrare: We have specs for this so we can always come back and improve this as we go

PR #259, opened 2024-08-09
Consistency cleanup

Fryguy:
@kbrock Please review.  This is just cleanup, splitting up specs into smaller units, and reordering.

- Break up failure specs into logical groups
  Also drop some redundant jsonpath tests, which are handled elsewhere.
- Split up JsonToString and StringToJson specs into individual tests
  Also
  - Add tests for both literal and jsonpath
  - Add tests for some missing cases for Float and false
- Ensure ordering matches states-language.net appendix B
- Add missing parser spec for JsonMerge
- Nicer format for StringToJson parsing error message

Follow up to #64

### DIFF
--- a/spec/workflow/intrinsic_function_spec.rb
+++ b/spec/workflow/intrinsic_function_spec.rb
@@ -589,28 +589,84 @@ RSpec.describe Floe::Workflow::IntrinsicFunction do
     end
 
     describe "States.JsonToString" do
-      it "fetches values from the input" do
-        # this is not in the spec but automatic as part of the parser
-        result = described_class.value("States.JsonToString(true)", {}, {})
+      it "with true" do
+        result = described_class.value("States.JsonToString(true)")
         expect(result).to eq("true")
+      end
 
-        result = described_class.value("States.JsonToString($.input)", {}, {"input" => "foo"})
-        expect(result).to eq("\"foo\"")
+      it "with false" do
+        result = described_class.value("States.JsonToString(false)")
+        expect(result).to eq("false")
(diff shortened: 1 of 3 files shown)


spec/workflow/intrinsic_function_spec.rb
```diff
@@ -589,30 +589,86 @@
     end
 
     describe "States.JsonToString" do
-      it "fetches values from the input" do
-        # this is not in the spec but automatic as part of the parser
```
Fryguy, line 563: Removed this comment because I checked all of these in the simulator and they all work with literals.

spec/workflow/intrinsic_function_spec.rb
```diff
@@ -583,14 +613,10 @@
         expect(result).to eq("[\"foo\",\"bar\"]")
       end
 
-      it "requires 1 parameter" do
+      it "fails with wrong number of arguments" do
         expect { described_class.value("States.JsonToString()", {}, {}) }.to raise_error(ArgumentError, "wrong number of arguments to States.JsonToString (given 0, expected 1)")
         expect { described_class.value("States.JsonToString($.input, true)", {}, {"input" => 5}) }.to raise_error(ArgumentError, "wrong number of arguments to States.JsonToString (given 2, expected 1)")
       end
-
-      it "requires a valid path" do
```
Fryguy, line 591: I dropped this test (and the other similar one), because we have a whole separate section in the tests for jsonpath failures making this redundant.

kbrock:
Nice Changes

PR #260, opened 2024-08-09
Enable support for Hashes in States.Hash

Fryguy:
While the actual hash value is different from the stepfunctions simulator it is still internally consistent, so it's usable. This would only create problems if someone ran this States.Hash in floe and tried to compare it to the exact same run in AWS, which seems unlikely. Even so, I added a pending test so we can try to figure it out later.

@kbrock or @agrare Please review.  Follow up to #64

### DIFF
--- a/lib/floe/workflow/intrinsic_function/transformer.rb
+++ b/lib/floe/workflow/intrinsic_function/transformer.rb
@@ -177,7 +177,7 @@ module Floe
         rule(:states_hash => {:args => subtree(:args)}) do
           args = Transformer.process_args(args(), "States.Hash", [Object, String])
           data, algorithm = *args
-          raise NotImplementedError if data.kind_of?(Hash)
+
           if data.nil?
             raise ArgumentError, "invalid value for argument 1 to States.Hash (given null, expected non-null)"
           end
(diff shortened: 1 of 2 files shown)


lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -177,7 +177,7 @@ def check_types(args, function, signature)
         rule(:states_hash => {:args => subtree(:args)}) do
           args = Transformer.process_args(args(), "States.Hash", [Object, String])
           data, algorithm = *args
-          raise NotImplementedError if data.kind_of?(Hash)
```
kbrock, line 180: Very nice

lib/floe/workflow/intrinsic_function/transformer.rb
```diff
@@ -189,8 +189,8 @@ def check_types(args, function, signature)
 
           require "openssl"
           algorithm = algorithm.sub("-", "")
-          data = data.to_json unless data.kind_of?(String)
-          OpenSSL::Digest.hexdigest(algorithm, data)
```
kbrock, line 193: Really don't like these `....json... if String` kind of statements

spec/workflow/intrinsic_function_spec.rb
```diff
@@ -501,7 +501,12 @@
       end
 
       it "with a hash" do
-        pending("Not implemented yet")
+        result = described_class.value("States.Hash($.data, 'SHA-1')", {}, {"data" => {"foo" => "bar"}})
+        expect(result).to eq("a5e744d0164540d33b1d7ea616c28f2fa97e754a")
```
kbrock, line 505: very cool you figured this one out

PR #263, opened 2024-08-12
Add step functions tool to gitignore

Fryguy:
Whether we merge #244 or not, I think it's useful to have a known place where the tool could exist, so this exacts the .gitignore entry from #244 for that purpose.  Locally this would help me a lot, cause every so often I accidentally commit it and it's huge :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@
 /tmp/
 
 Gemfile.lock
+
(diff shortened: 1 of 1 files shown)


Fryguy:
Closing since we discussed installing into a system location instead, and perhaps I might open up a homebrew PR for installation.

PR #265, opened 2024-08-14
Fix podman/docker container_ref trailing newline

agrare:
Since docker and podman use AwesomeSpawn to run a command-line utility the result that we use to get the container_ref has a trailing newline.

This causes issues trying to match the container_ref with one coming from events.

```
   131:         result = docker!(*params)
   132:         require 'byebug'; byebug
=> 133:         result.output
   134:       end

(byebug) result.output
"7c640ef968d2ce923418a3de8eedae5a19c0a31f1264417b0088e0df57d0bac8\n"
```

### DIFF
--- a/lib/floe/container_runner/docker.rb
+++ b/lib/floe/container_runner/docker.rb
@@ -129,7 +129,7 @@ module Floe
         logger.debug("Running #{AwesomeSpawn.build_command_line(self.class::DOCKER_COMMAND, params)}")
 
         result = docker!(*params)
-        result.output
+        result.output.chomp
       end
 
       def run_container_params(image, env, secrets_file)
--- a/spec/container_runner/docker_spec.rb
+++ b/spec/container_runner/docker_spec.rb
@@ -15,21 +15,27 @@ RSpec.describe Floe::ContainerRunner::Docker do
     end
 
     it "calls docker run with the image name" do
-      stub_good_run!("docker", :params => ["run", :detach, [:name, a_string_starting_with("floe-hello-world-")], "hello-world:latest",], :output => container_id)
+      stub_good_run!("docker", :params => ["run", :detach, [:name, a_string_starting_with("floe-hello-world-")], "hello-world:latest",], :output => "#{container_id}\n")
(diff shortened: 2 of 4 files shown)


lib/floe/container_runner/docker.rb
```diff
@@ -30,7 +30,8 @@ def run_async!(resource, env = {}, secrets = {}, _context = {})
         end
 
         begin
-          runner_context["container_ref"] = run_container(image, env, runner_context["secrets_ref"])
+          result = run_container(image, env, runner_context["secrets_ref"])
+          runner_context["container_ref"] = result.strip
```
Fryguy, line 34: I'm fine with the strip, but I wonder if run_container method should just do a `chomp` at the end regardless - that bonus newline might affect a bunch of things.

agrare: Yeah that's true I can do that instead

lib/floe/container_runner/docker.rb
```diff
@@ -102,6 +103,7 @@ def status!(runner_context)
         return if runner_context.key?("Error")
 
         runner_context["container_state"] = inspect_container(runner_context["container_ref"])&.dig("State")
+        runner_context["container_state"]
```
kbrock, line 106: I don't understand the need for this line

Fryguy: Same - the previous line returns that same value, so this is not needed.

agrare: My mistake I had some debug code in here to dump the state and this was left over, great catch

Fryguy (approved):

Fryguy:
Is it possible to add a spec for this?

kbrock:
> Is it possible to add a spec for this?

Heh. this is totally why I wanted to add the `AwesomeSpawn` runner. So we could use that provider in tests.

agrare:
@Fryguy @kbrock okay added tests covering this case

Fryguy (approved):

PR #278, opened 2024-09-19
Introduce WorkflowBase

kbrock:
# Goal

`WorkflowBase` is the root of both `Map` and `Parallel`.
Pull this concept out of #184

# Contents

This is the second commit from #184 by @agrare 

- Run `ItemProcessor` workflow for `Map` state
- Introduce `WorkflowBase` **<== this PR**
- ...

Branch [kbrock#add_map_state](https://github.com/kbrock/floe/tree/add_map_state) contains the whole branch, but with only the first two commits swapped. This may help with your local rebasing.

### DIFF
--- a/lib/floe.rb
+++ b/lib/floe.rb
@@ -8,6 +8,7 @@ require_relative "floe/logging"
 require_relative "floe/runner"
 
 require_relative "floe/validation_mixin"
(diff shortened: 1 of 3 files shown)


kbrock:
defer to #279

PR #282, opened 2024-10-07
Map state add tolerated failure

agrare:
Adds checking of child iterations of a Map state for failure, and based on ToleratedFailureCount/Percentage marks the Map state as failed or successful


lib/floe/workflow/states/task.rb
```diff
@@ -1,11 +1,16 @@
 # frozen_string_literal: true
 
+require_relative "input_output_mixin"
+require_relative "non_terminal_mixin"
+require_relative "retry_catch_mixin"
```
kbrock, line 5: Feels like these belong with the other `requires` in the global `floe.rb`

agrare: Yes, it is there as well but keeping those alphabetic means this fails to resolve the constant.
Alternatively, I can move all of the _mixin requires above the others in floe.rb

lib/floe/workflow/states/task.rb
```diff
@@ -82,54 +87,6 @@ def success?(context)
...
-          # If a different retrier is hit reset the context
-          if !context["State"].key?("RetryCount") || context["State"]["Retrier"] != retrier.error_equals
-            context["State"]["RetryCount"] = 0
-            context["State"]["Retrier"]    = retrier.error_equals
-          end
-
-          context["State"]["RetryCount"] += 1
-
-          return if context["State"]["RetryCount"] > retrier.max_attempts
-
```
kbrock, line 106: Always wanted the retrier / catcher to act just like a `State`.

You ask - do you have a retrier/catcher for me?
And if you do, then you `run_nonblock` it just like you do for the current state.

lib/floe/workflow/states/retry_catch_mixin.rb
```diff
@@ -0,0 +1,57 @@
...
+          true
+        end
+
+        def catch_error!(context, error)
+          catcher = find_catcher(error["Error"]) if error
+          return if catcher.nil?
+
+          context.next_state = catcher.next
+          context.output     = catcher.result_path.set(context.input, error)
+          logger.info("Running state: [#{long_name}] with input [#{context.json_input}]...CatchError - next state: [#{context.next_state}] output: [#{context.json_output}]")
```
kbrock, line 42: catcher and retrier always seemed like the same thing.

you try and match it, and if it matches, then you set the next_state / output

lib/floe/workflow/states/task.rb
kbrock: I like making the require not alphabetical.

agrare: I like the require being in the place that needs it and not trying to solve a dependency graph in floe.rb :laughing:

lib/floe/workflow/states/retry_catch_mixin.rb
agrare: They are similar for sure, not sure the same thing though

agrare:
@kbrock I don't want to get too bogged down by retrier / catcher changes, is there a way we can tackle that in a separate PR and for this one just say we're using catcher/retrier from Map and Task?  Maybe I even drop it from Map completely and we add it later.

lib/floe/workflow/states/map.rb
```diff
@@ -58,8 +60,13 @@ def start(context)
...
         def finish(context)
-          result = context.state["ItemProcessorContext"].map { |ctx| Context.new(ctx).output }
-          context.output = process_output(context, result)
+          if failed?(context)
+            error = parse_error(context)
+            retry_state!(context, error) || catch_error!(context, error) || fail_workflow!(context, error)
+          else
+            result = each_item_processor(context).map(&:output)
+            context.output = process_output(context, result)
+          end
```
kbrock, line 69: Please change to `success?()` to mirror `Task`

lib/floe/workflow/states/map.rb
```diff
@@ -77,27 +84,47 @@ def end?
...
-          context.state["ItemProcessorContext"].all? { |ctx| Context.new(ctx).ended? }
+          each_item_processor(context).all?(&:ended?)
+        end
+
+        def failed?(context)
+          contexts = each_item_processor(context)
+
+          # Handle the simple cases first
+          return true  if contexts.all?(&:failed?)
+          return false if contexts.none?(&:failed?)
```
kbrock, line 111: Probably for another PR, but feels like these 2 can be dropped.

Not sure what happens if 1 fails and `tolerated_failure_cont` / tolerated_failure_percentage` are not populated. Can we default `tolerated_failure_percentage = 0`?

If that is the case, then we can drop above 2 and all should work?

lib/floe/workflow/states/retry_catch_mixin.rb
kbrock: ignore ^

I can play with this refactor later

lib/floe/workflow/states/task.rb
kbrock: ignore ^ - I can play with this later

lib/floe/workflow/states/map.rb
Fryguy: +1 for preferring positively toned methods (avoids double-negatives and such)

lib/floe/workflow/states/map.rb
kbrock: probably punt here

lib/floe/workflow/states/task.rb
kbrock: huh.

We have been putting all requires up front in `floe.rb`.
(I did not just say: we have been doing it this way and we have to continue doing that)

Do we want to get away from that?

wondering if InputOutputMixin and the others do not belong in States.
That way we can just require these and then all states and not have to deal with orders.

Does look like these 3 mixins are used by only states

lib/floe/workflow/states/map.rb
```diff
@@ -77,27 +85,46 @@ def end?
...
+          contexts   = each_item_processor(context)
+          num_failed = contexts.count(&:failed?)
+          total      = contexts.count
+
+          return true if num_failed.zero?
+
+          # Some have failed, check the tolerated_failure thresholds to see if
+          # we should fail the whole state.
+          return true if tolerated_failure_count      && num_failed < tolerated_failure_count
+          return true if tolerated_failure_percentage && (100 * num_failed / total.to_f) < tolerated_failure_percentage
```
Fryguy, line 117: What is supposed to happen if both count and percentage are given (or is that not allowed)?  I can think of percentages where you want, say, the minimum of 50% or 3.  In that case, I think you need to check both clauses, and then return, as opposed to bailing out on the first one.

(This could have fallen through with the flip from failed? to success?)

kbrock: specs tend to say one vs the other.
In the branch that I had with all error checkings, I only allow or the other.

Also, you should not be able to state Next and End at the same time.
But in the short term, we've been letting these cases slide

Fryguy: From the states language spec:

> If a "ToleratedFailurePercentage" field and a "ToleratedFailureCount" field are both specified, the Map State will fail if either threshold is breached.

Fryguy: There's an additional nuance here that I'm wondering if we need to code explicitly (not 100% sure).  The spec says

> A Map State MAY have a "ToleratedFailurePercentage" field whose value MUST be a number between zero and 100. Its default value is zero, which means the Map State will fail if any (i.e. more than 0%) of its items fail. A "ToleratedFailurePercentage" value of 100 means the interpreter will continue starting iterations even if all items fail.

So if 0 or 100 are specified, then those have defined meanings, and I'm concerned that silly floating point math might let those fall through the cracks?  I wonder if we should have at least some tests for those specific values.

Fryguy: Another very strange question but is it possible for `total` to be 0 here or does some earlier check avoid that?  Asking because this is a potential divide by zero here.

agrare: > From the states language spec:
>> If a "ToleratedFailurePercentage" field and a "ToleratedFailureCount" field are both specified, the Map State will fail if either threshold is breached.

Right this will report a failure if either threshold is hit

agrare: > Another very strange question but is it possible for total to be 0 here or does some earlier check avoid that? Asking because this is a potential divide by zero here.

Technically this will return early if total is zero because num_failed will also be zero, but I can add an additional / explicit total.zero? above

Fryguy:
I noticed the spec also has `States.ExceedToleratedFailureThreshold` as an error code.  I'm not sure if/how we support these error codes, but just wanted to mention.

lib/floe/workflow/states/task.rb
agrare: > We have been putting all requires up front in floe.rb.
> (I did not just say: we have been doing it this way and we have to continue doing that)

No I don't think we need to stop doing that, I think of `require 'floe'` as "just require everything" but if a particular class needs something specific it should require it that way a caller could `require 'floe/workflow/context'` and not having to bring in the memory of the entire gem.

lib/floe/workflow/states/map.rb
Fryguy: > Right this will report a failure if either threshold is hit

The way it's coded though, I'm not sure it does?  Taking a (very) contrived example, if we had 4 items, 2 failures,  threshold count of 2, and the threshold % of 25%, then the current code will return success true, but should return false.

lib/floe/workflow/states/map.rb
agrare: I moved some things around so this is a little simpler now

lib/floe/workflow/states/map.rb
agrare: Okay I added an explicit check for ToleratedFailurePercentage==100 (interesting the spec says it is an integer not a float so that made `== 100` easier, I did `&.to_i` in the initialize)
I think all concerns here are convered.

agrare: > The way it's coded though, I'm not sure it does? Taking a (very) contrived example, if we had 4 items, 2 failures, threshold count of 2, and the threshold % of 25%, then the current code will return success true, but should return false.

Oh yeah, this did flip when going from `failed?` to `success?`

agrare:
> I noticed the spec also has States.ExceedToleratedFailureThreshold as an error code. I'm not sure if/how we support these error codes, but just wanted to mention.

Yeah I'm curious would this only be the error if `ToleratedFailureCount` or `ToleratedFailurePercentage` is present otherwise what would the error be?  I'm taking the error from the failed sub-workflows in this case.

PR #283, opened 2024-10-07
Run Map iterations in parallel up to MaxConcurrency

agrare:
Important to note that while it is possible for multiple iterations to be running in parallel externally (aka docker containers), there is still only a single thread checking for branch status so IMO we don't have any concurrency issues with the Context hash here.

### DIFF
--- a/examples/map.asl
+++ b/examples/map.asl
@@ -19,6 +19,7 @@
     "Map demo": {
       "Type": "Map",
       "ItemsPath": "$.colors",
+      "MaxConcurrency": 2,
       "ItemProcessor": {
         "ProcessorConfig": {
           "Mode": "INLINE"
@@ -27,11 +28,15 @@
         "States": {
           "Generate UUID": {
             "Type": "Pass",
-            "End": true,
+            "Next": "Sleep",
             "Parameters": {
               "uuid.$": "States.UUID()"
-            },
-            "OutputPath": "$.uuid"
+            }
+          },
+          "Sleep": {
+            "Type": "Task",
+            "Resource": "docker://docker.io/agrare/sleep:latest",
+            "End": true
           }
         }
       },
--- a/lib/floe/workflow/states/map.rb
+++ b/lib/floe/workflow/states/map.rb
@@ -68,7 +68,7 @@ module Floe
         def run_nonblock!(context)
           start(context) unless context.state_started?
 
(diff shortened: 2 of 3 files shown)


examples/map.asl
```diff
@@ -27,11 +28,15 @@
...
             "Parameters": {
               "uuid.$": "States.UUID()"
-            },
-            "OutputPath": "$.uuid"
+            }
+          },
+          "Sleep": {
+            "Type": "Task",
+            "Resource": "docker://docker.io/agrare/sleep:latest",
+            "End": true
```
agrare, line 39: NOTE this gives us a sub-iteration that runs a task that doesn't complete immediately like a Pass state would so it exercises the Workflow.wait event-handling code for child-tasks

spec/workflow/states/map_spec.rb
```diff
@@ -26,7 +26,7 @@
         "Type"           => "Map",
         "InputPath"      => "$.detail",
         "ItemsPath"      => "$.shipped",
-        "MaxConcurrency" => 0,
```
agrare, line 29: I was wondering why my spec was hanging :laughing: 
Add a check to raise a workflow error if MaxConcurrency was 0

agrare:
> Method step_nonblock! has a Cognitive Complexity of 6 (exceeds 5 allowed). Consider refactoring

Okay I got this down to 6 (I don't usually chase cognitive complexity but this method was a little gross and now is improved)

lib/floe/workflow/states/map.rb
```diff
@@ -74,7 +72,7 @@ def finish(context)
         def run_nonblock!(context)
           start(context) unless context.state_started?
 
-          loop while step_nonblock!(context) == 0 && running?(context)
+          step_nonblock!(context) while running?(context)
```
kbrock, line 75: is this a universal change you want to make?
(not asking for the change, just trying to understand trajectory)

agrare: Specifically for map this was now essentially doing the same running? check twice (ended? is just !running?) so it was more efficient to do this

agrare:
@kbrock can you take a look?

Fryguy:
Is Map + Max Concurrency different than Parallel?

kbrock (approved):
This has some great changes.

We did discuss possibly not looping the contexts for so many of the operations and using a hash for the contexts. (for future thought)

PR #284, opened 2024-10-08
Handle either tolerated_failure_count or tolerated_failure_percentage

agrare:
Follow-up to https://github.com/ManageIQ/floe/pull/282#discussion_r1792238478 to handle both ToleratedFailureCount and ToleratedFailurePercent present

### DIFF
--- a/lib/floe/workflow/states/map.rb
+++ b/lib/floe/workflow/states/map.rb
@@ -110,13 +110,18 @@ module Floe
           total      = contexts.count
 
           return true if num_failed.zero? || total.zero?
-          return true if tolerated_failure_percentage && tolerated_failure_percentage == 100
+          return false if tolerated_failure_count.nil? && tolerated_failure_percentage.nil?
+
           # Some have failed, check the tolerated_failure thresholds to see if
           # we should fail the whole state.
-          return true if tolerated_failure_count      && num_failed < tolerated_failure_count
-          return true if tolerated_failure_percentage && (100 * num_failed / total.to_f) < tolerated_failure_percentage
+          #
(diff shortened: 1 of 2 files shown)


spec/workflow/states/map_spec.rb
```diff
@@ -263,6 +263,15 @@
           end
         end
       end
+
+      context "with both ToleratedFailureCount and ToleratedFailurePercentage" do
+        let(:tolerated_failure_percentage) { 50 }
+        let(:tolerated_failure_count)      { 1 }
```
Fryguy, line 269: Can you add another spec showing when count might be tolerated but not percentage (i.e. the flipped case to this?)

Fryguy:
My brain is having trouble with ||= with booleans for some reason, but I think this is good.

spec/workflow/states/map_spec.rb
agrare: :+1: done, also added a case where they are both present but neither high enough for state success

agrare:
> My brain is having trouble with ||= with booleans for some reason, but I think this is good.

Yeah it is funny how used to `||=` being for `nil` we get :smile:

Fryguy (approved):

lib/floe/workflow/states/map.rb
```diff
@@ -110,13 +110,18 @@ def success?(context)
           total      = contexts.count
 
           return true if num_failed.zero? || total.zero?
-          return true if tolerated_failure_percentage && tolerated_failure_percentage == 100
+          return false if tolerated_failure_count.nil? && tolerated_failure_percentage.nil?
```
kbrock, line 113: can this be a configuration error and not a runtime error?

PR #285, opened 2024-10-08
Return ExceedToleratedFailureThreshold if ToleratedFailureCount/Percentage is present

agrare:
Follow-up to https://github.com/ManageIQ/floe/pull/282#issuecomment-2400408415

### DIFF
--- a/lib/floe/workflow/states/map.rb
+++ b/lib/floe/workflow/states/map.rb
@@ -142,7 +142,14 @@ module Floe
         end
 
         def parse_error(context)
-          each_item_processor(context).detect(&:failed?)&.output&.dig("Error")
+          # If ToleratedFailureCount or ToleratedFailurePercentage is present
+          # then use States.ExceedToleratedFailureThreshold otherwise
+          # take the error from the first failed state
+          if tolerated_failure_count || tolerated_failure_percentage
+            {"Error" => "States.ExceedToleratedFailureThreshold"}
+          else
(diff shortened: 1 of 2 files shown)


kbrock:
This is an improvement.

For all States, we store the error in the current's output.
Can we do that too?

This way, the js front end won't need all this logic.
And when we come back in, we won't need to re-de-serialize all this data.

agrare:
> For all States, we store the error in the context.state.output.
> Can we do this too after the task is "done"?

Actually this is exactly what is being done, e.g. see the specs: https://github.com/ManageIQ/floe/pull/285/files#diff-926c9da5ddabfe46a43d14e36a588fff096e8b76bbd56415802f84597a6a1405R169

The error is stored the exact same way as Task is currently doing it.

>  Maybe store the number of failures in the context as well?

Not a bad idea, but I'm thinking that the very next thing a user would want to see is the status of each of the iterations also with a pretty icon and a list of inputs/outputs and the error message anyway

kbrock:
> > For all States, we store the error in the context.state.output.
> > Can we do this too after the task is "done"?
> 
> Actually this is exactly what is being done,

```ruby
# success?
num_failed = contexts.count(&:failed?)

# parse_error
each_item_processor(context).detect(&:failed?)&.output || {"Error" => "States.Error"}
```

I was hoping to only see one iterator counting failed.

PR #288, opened 2024-10-14
Consolidate Mixin requires

kbrock:
@agrare Does this solve your require issue?

Or would we need to split out the requires for each level?
and put requires in floe.rb and workflow.rb and ...


agrare (approved):

agrare:
I still don't see the problem with adding requires in the files in which they're used, but this does also work

PR #289, opened 2024-10-16
Extract methods cli

kbrock:
extracted from https://github.com/ManageIQ/floe/pull/223

`Cli#run` method is a bit complicated, so extracting these methods to trim it down

### DIFF
--- a/lib/floe/cli.rb
+++ b/lib/floe/cli.rb
@@ -22,8 +22,7 @@ module Floe
 
       workflows =
         workflows_inputs.each_slice(2).map do |workflow, input|
-          context = Floe::Workflow::Context.new(opts[:context], :input => input, :credentials => credentials)
-          Floe::Workflow.load(workflow, context)
+          create_workflow(workflow, opts[:context], input, credentials)
         end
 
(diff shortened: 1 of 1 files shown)


lib/floe/cli.rb
```diff
@@ -82,5 +76,20 @@ def parse_options!(args)
...
       return workflows_inputs, opts
     end
+
+    def create_credentials(opts)
+      if opts[:credentials_given]
+        opts[:credentials] == "-" ? $stdin.read : opts[:credentials]
+      elsif opts[:credentials_file_given]
+        File.read(opts[:credentials_file])
+      else
+        {}
```
agrare, line 86: @kbrock this looks like a functional change (nil vs {} when credentials not passed).  Was this intentional?  Not a big deal since in Context we do `@credentials = credentials || {}` anyway just curious why the change

kbrock: I'm overly sensitive right now to us making conversions (e.g.: `making JSON.parse(x) if x.kind_of?(String)` )

I'll undo this. Though I kinda like it better this way.

agrare (approved):

Issue #241, opened 2024-07-03
Support Map/Parallel


kbrock:
Add support for Map and Parallel.
These are the last remaining States to be implemented

PR #294, opened 2024-11-05
Add Map ItemBatcher/ItemSelector

agrare:
The ItemBatcher provides a way to group large array inputs into larger batches for better performance

You can provide a `MaxItemsPerBatch` or `MaxItemsPerBatchPath` which limits the number of items in a batch, as well as a `MaxInputBytesPerBatch` and `MaxInputBytesPerBatchPath` which limits the size of the input in bytes.

The output will be a set of `{"Items": []}` payloads, one for each batch.

> Batching Items
>
> A Map State MAY have an "ItemBatcher" field, whose value MUST be a JSON object and is called the ItemBatcher Configuration. The ItemBatcher Configuration causes the interpreter to batch selected items into sub-arrays before passing them to each invocation. The interpreter will limit each sub-array to the maximum number of items specified by the "MaxItemsPerBatch" field, and to the maximum size in bytes specified by the "MaxInputBytesPerBatch" field.
>
> The ItemBatcher Configuration MAY have a "BatchInput" field, whose value MUST be a Payload Template. An ItemBatcher Configuration MAY have a "MaxItemsPerBatch" field, whose value MUST be a positive integer. An ItemBatcher Configuration MAY have a "MaxInputBytesPerBatch" field, whose value MUST be a positive integer.
>
> The default of "ItemBatcher" is the selected item. Put another way, the interpreter will not batch items if no "ItemBatcher" field is provided.
>
> Both the "MaxItemsPerBatch" and "MaxInputBytesPerBatch" can be provided indirectly. A Map State may have "MaxItemsPerBatchPath" and "MaxInputBytesPerBatchPath" fields which MUST be Reference Paths which, when resolved, MUST select fields whose values are positive integers. A Map State MUST NOT include both "MaxItemsPerBatch" and "MaxItemsPerBatchPath" or both "MaxInputBytesPerBatch" and "MaxInputBytesPerBatchPath".
>
> An ItemBatcher Configuration MUST contain at least one of "MaxItemsPerBatch", "MaxItemsPerBatchPath", "MaxInputBytesPerBatch", or "MaxInputBytesPerBatchPath".

https://docs.aws.amazon.com/step-functions/latest/dg/input-output-itembatcher.html
https://github.com/ManageIQ/floe/issues/241


lib/floe/workflow/item_batcher.rb
```diff
@@ -0,0 +1,44 @@
...
+        parser_error!("must not specify both \"MaxInputBytesPerBatch\" and \"MaxInputBytesPerBatchPath\"")     if max_input_bytes_per_batch && max_input_bytes_per_batch_path
+        invalid_field_error!("MaxItemsPerBatch", max_items_per_batch, "must be a positive integer")            if max_items_per_batch && max_items_per_batch <= 0
+        invalid_field_error!("MaxInputBytesPerBatch", max_input_bytes_per_batch, "must be a positive integer") if max_input_bytes_per_batch && max_input_bytes_per_batch <= 0
+      end
+
+      def value(context, input, state_input = nil)
+        state_input ||= input
+
+        max_items       = max_items_per_batch       || max_items_per_batch_path&.value(context, state_input)
+        max_input_bytes = max_input_bytes_per_batch || max_input_bytes_per_batch_path&.value(context, state_input)
```
kbrock, line 34: codeclimate: this is not used.

agrare: Correct, I haven't implemented it yet :D

agrare: NOTE I also need to check for a runtime exception here if they used a Path and the value was `<= 0`

lib/floe/workflow/item_batcher.rb
```diff
@@ -0,0 +1,57 @@
...
+        return if max_items_per_batch_path.nil?
+
+        result = max_items_per_batch_path.value(context, state_input)
+        raise runtime_field_error!("MaxItemsPerBatchPath", result, "must be a positive integer") if result <= 0
+
+        result
+      end
+
+      def validate!
+        if [max_items_per_batch, max_items_per_batch_path].all?(&:nil?)
```
Fryguy, line 46: Should this be any? if you only need one of them?

```suggestion
        if [max_items_per_batch, max_items_per_batch_path].any?(&:nil?)
```

lib/floe/workflow/item_batcher.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+
+      def value(context, input, state_input = nil)
+        state_input ||= input
+
+        output = batch_input ? batch_input.value(context, state_input) : {}
+
+        input.each_slice(max_items(context, state_input)).map do |batch|
+          output.merge("Items" => batch)
+        end
```
kbrock, line 30: future: would be nice to not resolve all input at the start.

lib/floe/workflow/item_batcher.rb
kbrock, line 46: optimization (read: feel free to ignore)
Count these and compare to 1?

lib/floe/workflow/item_batcher.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      def validate!
+        if [max_items_per_batch, max_items_per_batch_path].all?(&:nil?)
+          parser_error!("must have one of \"MaxItemsPerBatch\", \"MaxItemsPerBatchPath\"")
+        end
+
+        parser_error!("must not specify both \"MaxItemsPerBatch\" and \"MaxItemsPerBatchPath\"")               if max_items_per_batch && max_items_per_batch_path
+        parser_error!("must not specify both \"MaxInputBytesPerBatch\" and \"MaxInputBytesPerBatchPath\"")     if max_input_bytes_per_batch && max_input_bytes_per_batch_path
+        invalid_field_error!("MaxItemsPerBatch", max_items_per_batch, "must be a positive integer")            if max_items_per_batch && max_items_per_batch <= 0
+        invalid_field_error!("MaxInputBytesPerBatch", max_input_bytes_per_batch, "must be a positive integer") if max_input_bytes_per_batch && max_input_bytes_per_batch <= 0
```
kbrock, line 53: not sure if we care, but these will throw a ruby runtime error if these are not an integer. I really want to avoid that. Maybe do a `to_i`?

spec/workflow/item_batcher_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+        expect(subject.value(context, input, state_input)).to eq([{"Items" => %w[a b]}, {"Items" => %w[c d]}, {"Items" => %w[e]}])
+      end
+
+      context "with an invalid value in input" do
+        let(:state_input) { {"batchSize" => 0, "items" => input} }
+
+        it "raises an exception" do
+          expect { subject.value(context, input, state_input) }
+            .to raise_error(Floe::ExecutionError, "Map.ItemBatcher field \"MaxItemsPerBatchPath\" value \"0\" must be a positive integer")
+        end
```
kbrock, line 149: Not sure if we want a test with invalid values. (are we doing tests for these?)

e.g.:

```ruby
let(:state_input) { {"batchSize" => "a", "items" => input} }
```

lib/floe/workflow/item_batcher.rb
agrare: You mean like `[max_items_per_batch, max_items_per_batch_path].compact.count.zero?`

lib/floe/workflow/item_batcher.rb
agrare: I can add a `kind_of?(Integer)` check, I don't want to "blindly" to_i it because that would hide someone passing in a float which is kind of unexpected.

agrare:
Added Integer type checks and tests for floats and strings

lib/floe/workflow/item_batcher.rb
kbrock: yea

```ruby
unless [max_items_per_batch, max_items_per_batch_path].compact.count == 1
  parser_error!("must have one of \"MaxItemsPerBatch\", \"MaxItemsPerBatchPath\"")
end
```

kbrock (approved):

PR #302, opened 2025-06-16
Add ASL example with all states

agrare:
Basically a merge of the basic "examples/workflow.asl" which had most states, plus map and parallel.

### DIFF
--- a/examples/everything.asl
+++ b/examples/everything.asl
@@ -1,0 +1,146 @@
+{
+  "Comment": "An example of the Amazon States Language with all states.",
+  "StartAt": "FirstState",
+  "States": {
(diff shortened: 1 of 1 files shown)


examples/everything.asl
```diff
@@ -0,0 +1,146 @@
+{
+  "Comment": "An example of the Amazon States Language using a choice state.",
+  "StartAt": "FirstState",
+  "States": {
+    "FirstState": {
+      "Type": "Task",
+      "Resource": "docker://docker.io/agrare/hello-world:latest",
```
Fryguy, line 7: Not for this PR, but can we get this into the manageiq org space?

PR #307, opened 2025-08-12
Fix credentials when passed via cli

agrare:
When passing a credentials payload via the CLI the context was built properly but the credentials field was being lost because of the default `{}` in Workflow.load

They also weren't parsed as a JSON payload so a string payload would end up as `@credentials = '{"password": "1234"}'` rather than a hash which could be accessed via JSONPath


kbrock:
I was really hoping to get away from constructors being responsible for serialization and deserialization.

I feel the CLI class should convert from a json string to actual ruby objects.
Or a class that reads/writes to the database should do the serialization and deserialization.

Changing this default value here to `nil` or `{}` doesn't do anything as the same logic is done just a few lines down `|| {}`

PR #308, opened 2025-08-12
Allow credentials referenced from parameters

agrare:
Built on:
- [x] https://github.com/ManageIQ/floe/pull/307

This allows for Credentials to be accessed from Parameters, e.g.:
```
"Headers.$": {
  "ContentType": "application/json",
  "Authorization.$": "States.Format('Basic {}', States.Base64Encode(States.Format('{}:{}', $$.Credentials.username, $$.Credentials.password)))"
}
```

NOTE I changed how the Task.Credentials field is processed so that it can be more consistent with how Credentials are handled by Parameters, that being `$.` takes from state input and `$$.Credentials` takes from Context.Credentials.

We'll have to update our examples to match this.

TODO:
- [x] Currently assigning to Credentials is special-cased in process_output checking for `ResultPath: $.Credentials`, but `$$.Credentials` blows up the Path initialize

### DIFF
--- a/lib/floe/workflow/context.rb
+++ b/lib/floe/workflow/context.rb
@@ -5,8 +5,6 @@ module Floe
     class Context
       include Logging
 
-      attr_accessor :credentials
-
       # @param context [Json|Hash] (default, create another with input and execution params)
       # @param input [Hash] (default: {})
       def initialize(context = nil, input: nil, credentials: nil, logger: nil)
@@ -14,14 +12,13 @@ module Floe
         input   = JSON.parse(input || "{}")
 
         @context = context || {}
+        self["Credentials"]        ||= credentials || {}
         self["Execution"]          ||= {}
         self["Execution"]["Input"] ||= input
         self["State"]              ||= {}
         self["StateHistory"]       ||= []
         self["StateMachine"]       ||= {}
         self["Task"]               ||= {}
-
-        @credentials = credentials || {}
 
         self.logger = logger if logger
       rescue JSON::ParserError => err
@@ -30,6 +27,10 @@ module Floe
 
       def execution
         @context["Execution"]
+      end
+
+      def credentials
+        @context["Credentials"]
       end
 
       def started?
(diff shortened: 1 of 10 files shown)


lib/floe/workflow/states/input_output_mixin.rb
```diff
@@ -15,9 +15,8 @@ def process_output(context, results)
           return if output_path.nil?
 
           results = result_selector.value(context, results) if @result_selector
-          if result_path.payload.start_with?("$.Credentials")
-            credentials = result_path.set(context.credentials, results)["Credentials"]
-            context.credentials.merge!(credentials)
+          if result_path.payload.start_with?("$$.Credentials")
```
Fryguy, line 18: Very minor, but I think we need to be a bit more pedantic here.  Technically it mattered with the old code more, but if someone had something liek `$.CredentialsStore`, this would have previously triggered and shouldn't have.  Think this should be something like.

```suggestion
          if result_path.payload.match?(/^$$.Credentials\b/)
```

kbrock:
huh. `except` must have been in active support and then moved to ruby in 3.0

lib/floe/workflow/context.rb
```diff
@@ -139,7 +140,7 @@ def dig(*args)
       end
 
       def to_h
-        @context
+        @context.except("Credentials")
```
Fryguy, line 143: I thikn we're also going to want to avoid printing Credentials in `.inspect` (because it's part of the `@context` ivar). That is, Context.new.inspect will print Credentials and we should hide that.  I have code I can share that will do that.

lib/floe/workflow/states/input_output_mixin.rb
agrare: So you can't normally assign anything to the Context object, we're making an exception here for `$$.Credentials` which means it isn't possible for someone to have `"ResultPath": "$$.CredentialsStore"` and have it be valid (this would fail previously but I've added a better validation that prints a helpful error message.)

lib/floe/workflow/context.rb
agrare: Done

agrare: :+1: sure thing, done

PR #317, opened 2025-09-05
add equality comparison (for specs)

kbrock:

### DIFF
--- a/lib/floe/workflow/context.rb
+++ b/lib/floe/workflow/context.rb
@@ -147,6 +147,10 @@ module Floe
         safe_context
       end
 
+      def ===(other)
+        other.is_a?(self.class) && other.instance_variable_get("@context") == @context
+      end
+
       private
 
       def safe_context
--- a/spec/workflow/context_spec.rb
+++ b/spec/workflow/context_spec.rb
@@ -216,5 +216,25 @@ RSpec.describe Floe::Workflow::Context do
         expect(ctx.to_h).not_to include("Credentials")
       end
     end
+
+    # NOTE: only used in specs
+    describe "===" do
+      let(:credentials) { {"username" => "user", "password" => "password"} }
+
+      it "equates values" do
+        ctx1 = described_class.new(:input => input.to_json)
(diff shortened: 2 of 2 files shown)


kbrock:
update:
- added == AND eq?
- updated spec typo (just introduced)

spec/workflow/context_spec.rb
```diff
@@ -214,5 +214,38 @@
     it "doesn't expose credentials" do
       expect(ctx_creds.to_h).not_to include("Credentials")
     end
+
+    # NOTE: introduced for specs
+    describe "==", "===" do
```
Fryguy, line 219: Please add .eql? and .hash tests as well.

lib/floe/workflow/context.rb
```diff
@@ -147,6 +147,11 @@ def to_h
         safe_context
       end
 
+      def ==(other)
+        other.is_a?(self.class) && other.instance_variable_get("@context") == @context
+      end
+      alias eql? ==
```
Fryguy, line 153: Pedantic, but if you override `.eql?` then that implies that `.hash` is the same, so you also need to override `.hash`.  Alternatively, I think if you override `.hash`, then `.eql?` works automatically and you don't need it.

spec/workflow/context_spec.rb
```diff
@@ -214,5 +214,38 @@
     it "doesn't expose credentials" do
       expect(ctx_creds.to_h).not_to include("Credentials")
     end
+
+    # NOTE: introduced for specs
```
agrare, line 218: I don't think we need to say why we introduced it, being able to compare `Context` can be helpful outside of just specs

lib/floe/workflow/context.rb
kbrock: I did add `hash`, but thought that was too much.

Interestingly, implementing `hash` does not fix `eql?`

kbrock:
update:
- added `#eql?` and `#hash` tests.
- removed `#===` tests (we did not add the method)
- removed comment around why we added the code

PR #321, opened 2025-09-17
Only workflow accesses workflow#payload

kbrock:
This is only used at `initialization` time.

This essentially makes `workflow#payload` private to `Workflow`/

Took out of #320 - the rest of that PR is not necessary

### DIFF
--- a/lib/floe/validation_mixin.rb
+++ b/lib/floe/validation_mixin.rb
@@ -23,7 +23,7 @@ module Floe
     end
 
     def workflow_state?(field_value, workflow)
-      workflow.payload["States"] ? workflow.payload["States"].include?(field_value) : true
+      workflow.workflow_state?(field_value, workflow)
     end
 
     def wrap_parser_error(field_name, field_value)
--- a/lib/floe/workflow_base.rb
+++ b/lib/floe/workflow_base.rb
@@ -77,6 +77,11 @@ module Floe
       context.output.to_json if end?(context)
     end
 
+    # overrides ValidationMixin#workflow_state?
+    def workflow_state?(field_value, _workflow)
+      @payload["States"].include?(field_value)
+    end
+
     private
 
     def step!(context)


lib/floe/validation_mixin.rb
```diff
@@ -23,7 +23,7 @@ def runtime_field_error!(field_name, field_value, comment, floe_error: "States.R
     end
 
     def workflow_state?(field_value, workflow)
-      workflow.payload["States"] ? workflow.payload["States"].include?(field_value) : true
+      workflow.workflow_state?(field_value, workflow)
```
agrare, line 26: What if instead of `workflow.payload["States"]` we use `workflow.states_by_name.key?(field_value)`

kbrock:
Also, it strikes me funny that `Workflow#workflow_state?` overrides `ValidationMixin#workflow_state?`. Feel like I'm doing something wrong.

kbrock:
update:
- rebase (no change. just wanted linear history)

kbrock:
update:
- using `names_for_state` when possible

This is used when we are parsing choices and children. TBH/ not sure why this is ever populated during validation. (validation happens at parse time. and we are recursively parsing the payload before `@states` and then `@names_for_state` are populated.

But made the change

lib/floe/workflow_base.rb
```diff
@@ -77,6 +77,11 @@ def output(context)
       context.output.to_json if end?(context)
     end
 
+    # overrides ValidationMixin#workflow_state?
+    def workflow_state?(field_value, _workflow)
+      @states_by_name ? @states_by_name.key?(field_value) : @payload["States"].include?(field_value)
```
agrare, line 82: Hm if we need to fallback to `@payload` anyway we should probably just use that

agrare:
> This is used when we are parsing choices and children. TBH/ not sure why this is ever populated during validation. (validation happens at parse time. and we are recursively parsing the payload before @states and then @names_for_state are populated.
>
> But made the change

Yeah I was hoping we could use that instead, doesn't seem to buy us anything if we have to fallback to `@payload` anyway

lib/floe/workflow_base.rb
Fryguy: yeah isn't this the same?
```suggestion
      @payload["States"].include?(field_value)
```

kbrock:
update:
- rebase (no conflict)
- just accessing payload directly

lib/floe/workflow_base.rb
```diff
@@ -77,6 +77,11 @@ def output(context)
       context.output.to_json if end?(context)
     end
 
+    # overrides ValidationMixin#workflow_state?
+    def workflow_state?(field_value, _workflow)
+      @payload["States"].include?(field_value)
```
Fryguy, line 82: Is it possible to get here and States not exist? - I'm not sure if the presence of States is validated first or not. If not, 
```suggestion
      @payload["States"]?.include?(field_value)
```

Fryguy (approved):
LGTM also want @agrare to review.

lib/floe/validation_mixin.rb
agrare: Do we need to pass in `workflow` as an argument if we're explicitly calling this on the `workflow` instance?  Seems redundant.

agrare: It seemed easier to throw up a PR to look at the diff, what do you think about https://github.com/ManageIQ/floe/pull/326/files ?

kbrock:
prefer #326

PR #343, opened 2026-06-16
Return an Error for failed http requests

agrare:
If an HTTP request fails we should return an error from the state. This extends the error payload with additional parameters including the status and body/headers.

Example unauthenticated API request:
```
irb(main):001> Floe::BuiltinRunner::Methods.http({"Url" => "http://localhost:3000/api/vms", "Options" => {"Encoding" => "JSON"}}, {}, {})
=> 
{"running"=>false,
 "success"=>false,
 "output"=>
  {"Error"=>"States.TaskFailed",
   "Cause"=>"Unauthorized"}
   "Details"=>
    {"Status"=>401,
     "Body"=>
      {"error"=>
        {"kind"=>"unauthorized",
         "message"=>"Api::BaseController::Authentication::AuthenticationError",
         "klass"=>"Api::BaseController::Authentication::AuthenticationError"}},
     "Headers"=>
      {"content-type"=>"application/json",
       "content-length"=>"169"}}}
}
```

After this change, the `http` builtin method will return an error payload for all non-200 return codes.
```json
{
  "Error": "States.Http.StatusCode.404",
  "Cause": "Not Found",
  "Details": {"Status": 404, "Body": {"code": 404, "reason": "Page not found"}, "Headers": {"content-type": "application/json"}}
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,27 @@ HTTP Parameters:
   * `User` - String - User for the proxy.
   * `Password` - String - Pasword for the proxy
 
+Output:
+A successful HTTP request will return the response Status, Body, and Headers. For example,
+```json
+{
+  "Status": 200,
+  "Body": {"result": {"foo": "bar"}},
+  "Headers": {"content-type": "application/json"}
+}
+```
+
+Errors:
+If an HTTP request fails, the `floe://http` builtin method will return an error payload that includes
+the error code, the reason, and the response Status, Body, and Headers. For example,
+```json
+{
+  "Error": "States.Http.StatusCode.404",
+  "Cause": "Not Found",
+  "Details": {"Status": 404, "Body": {"code": 404, "reason": "Page not found"}, "Headers": {"content-type": "application/json"}}
+}
+```
+
 #### Docker resource
 
 The docker resource runner takes a docker image URI, including the registry, name, and tag.
--- a/lib/floe/builtin_runner.rb
+++ b/lib/floe/builtin_runner.rb
@@ -7,9 +7,11 @@ module Floe
     SCHEME_PREFIX = "#{SCHEME}://".freeze
 
     class << self
-      def error!(runner_context = {}, cause:, error: "States.TaskFailed")
+      def error!(runner_context = {}, cause: nil, error: "States.TaskFailed", output: nil)
+        output ||= {"Error" => error, "Cause" => cause}
+
         runner_context.merge!(
-          "running" => false, "success" => false, "output" => {"Error" => error, "Cause" => cause}
+          "running" => false, "success" => false, "output" => output
         )
       end
 
--- a/lib/floe/builtin_runner.rb
+++ b/lib/floe/builtin_runner.rb
@@ -7,8 +7,8 @@ module Floe
     SCHEME_PREFIX = "#{SCHEME}://".freeze
 
     class << self
-      def error!(runner_context = {}, cause: nil, error: "States.TaskFailed", output: nil)
-        output ||= {"Error" => error, "Cause" => cause}
+      def error!(runner_context = {}, cause: nil, error: "States.TaskFailed", details: nil)
+        output = {"Error" => error, "Cause" => cause, "Details" => details}.compact
 
         runner_context.merge!(
           "running" => false, "success" => false, "output" => output
--- a/lib/floe/builtin_runner/methods.rb
+++ b/lib/floe/builtin_runner/methods.rb
@@ -92,13 +92,12 @@ module Floe
           faraday_request.body = body if body
         end
 
-        output = {"Status" => response.status, "Body" => response.body, "Headers" => response.headers}
-
         if response.success?
+          output = {"Status" => response.status, "Body" => response.body, "Headers" => response.headers}
           BuiltinRunner.success!({}, :output => output)
         else
-          output["Error"] = "States.TaskFailed"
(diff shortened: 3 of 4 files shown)


lib/floe/builtin_runner/methods.rb
```diff
@@ -94,7 +94,16 @@ def self.http(params, _secrets, _context)
 
         output = {"Status" => response.status, "Body" => response.body, "Headers" => response.headers}
 
-        BuiltinRunner.success!({}, :output => output)
+        if response.success?
+          BuiltinRunner.success!({}, :output => output)
+        else
+          output["Error"] = "States.TaskFailed"
```
agrare: Custom HTTP ASL errors would be cool here, `HTTP.NotFound` or something like that.  Would make retry matchers simpler

https://states-language.net/#errors
> States MAY report errors with other names, which MUST NOT begin with the prefix "States.".

agrare: Per https://docs.aws.amazon.com/step-functions/latest/dg/call-https-apis.html the aws-stepfunctions http lambda returns e.g. `States.Http.StatusCode.404`

Fryguy:
As discussed, I think it would be better to have a single top-level key `"Details"`, which can then have whatever we want underneath. Also, we should try to upstream the proposal to the spec.

lib/floe/builtin_runner/methods.rb
```diff
@@ -92,9 +92,16 @@ def self.http(params, _secrets, _context)
           faraday_request.body = body if body
         end
 
-        output = {"Status" => response.status, "Body" => response.body, "Headers" => response.headers}
+        if response.success?
+          output = {"Status" => response.status, "Body" => response.body, "Headers" => response.headers}
+          BuiltinRunner.success!({}, :output => output)
+        else
+          output = {"Error" => "States.Http.StatusCode.#{response.status}", "Cause" => response.reason_phrase, "Details" => response.body}
```
Fryguy: Oh I was thinking the details would have the entire output, including the response header and response status (it's ok if it's duplicated)


```suggestion
          output = {"Error" => "States.Http.StatusCode.#{response.status}", "Cause" => response.reason_phrase, "Details" => output}
```

lib/floe/builtin_runner.rb
```diff
@@ -7,9 +7,11 @@ module BuiltinRunner
     SCHEME_PREFIX = "#{SCHEME}://".freeze
 
     class << self
-      def error!(runner_context = {}, cause:, error: "States.TaskFailed")
+      def error!(runner_context = {}, cause: nil, error: "States.TaskFailed", output: nil)
+        output ||= {"Error" => error, "Cause" => cause}
```
Fryguy: I'm not sure I like the incoming output to have the error and the cause. I wonder if we should just add a details parameter here that's optional.


```suggestion
        output = {"Error" => error, "Cause" => cause}
        output["Details"] = details if details
```

Fryguy: Or if we're ok on Cause also being optional then something like


```suggestion
        output = {"Error" => error, "Cause" => cause, "Details" => details}.compact
```

lib/floe/builtin_runner/methods.rb
agrare: :+1: done

lib/floe/builtin_runner.rb
agrare: Agreed if we are limiting the error payload keys to only Error, Cause, and Details we don't need an arbitrary "output" parameter here

Fryguy:
PR title and body needs an update now too.

lib/floe/builtin_runner.rb
```diff
@@ -7,9 +7,13 @@ module BuiltinRunner
     SCHEME_PREFIX = "#{SCHEME}://".freeze
 
     class << self
-      def error!(runner_context = {}, cause:, error: "States.TaskFailed")
+      def error!(runner_context = {}, cause: nil, error: "States.TaskFailed", details: nil)
+        output = {"Error" => error}
+        output["Cause"]   = cause if cause
+        output["Details"] = details if details
```
Fryguy: Unless .compact is not available on all Rubies we support?

```suggestion
        output = {"Error" => error, "Cause" => cause, "Details" => details}.compact
```

README.md
```diff
@@ -246,6 +246,29 @@ HTTP Parameters:
   * `User` - String - User for the proxy.
   * `Password` - String - Pasword for the proxy
 
+Output:
+A successful HTTP request will return the Status, Body, and Headers:
```
Fryguy:
```suggestion
A successful HTTP request will return the response Status, Body, and Headers. For example,
```

README.md
```diff
@@ -246,6 +246,29 @@ HTTP Parameters:
...
+  "Body": {"result": {"foo": "bar"}},
+  "Headers": {"content-type": "application/json"}
+}
+```
+
+Errors:
+If an HTTP request fails the `floe://http` bultin method will return an error payload that includes
+the error code, the reason, and the response body.
+
+For example a 404 error might look like:
```
Fryguy:
```suggestion
If an HTTP request fails, the `floe://http` builtin method will return an error payload that includes
the error code, the reason, and the response Status, Body, and Headers. For example,
```

Fryguy (approved):

PR #344, opened 2026-06-17
`kill -TERM` the docker wait process before exiting

agrare:
When exiting the wait loop we should ensure that the docker/podman wait process is terminated.

This prevents `podman events` processes from hanging around after the caller has exited.  This isn't an issue on systems using systemd because the entire worker slice is torn down, but on development machines these will continue to run after the caller has exited.

Related:
- [ ] https://github.com/ManageIQ/awesome_spawn/pull/85

### DIFF
--- a/lib/floe/container_runner/docker.rb
+++ b/lib/floe/container_runner/docker.rb
@@ -92,10 +92,12 @@ module Floe
           Process.kill(0, pid)
         rescue Errno::ESRCH
           # Break out of the loop if the `docker events` process has exited
+          pid = nil
           break
         end
       ensure
         r.close
+        sigterm(pid) if pid
       end
 
       def status!(runner_context)


lib/floe/container_runner/docker.rb
```diff
@@ -92,10 +92,16 @@ def wait(timeout: nil, events: %i[create update delete], &block)
           Process.kill(0, pid)
         rescue Errno::ESRCH
           # Break out of the loop if the `docker events` process has exited
+          pid = nil
           break
         end
       ensure
         r.close
+        begin
+          Process.kill("TERM", pid) if pid
```
Fryguy: Should this use the `sigterm` method like we do on line 86?

agrare: I forgot we had that, yes it should thanks

Fryguy (approved):

PR #345, opened 2026-06-18
Use Kernel.spawn directly to ensure no intermediate `/bin/sh -c` shell is used

agrare:
I'm working on an update to AwesomeSpawn (https://github.com/ManageIQ/awesome_spawn/pull/85) but since AwesomeSpwan handles way more cases then we need for this purpose I thought this would be a quick way to fix the issue of `podman events` sticking around after exit.

### DIFF
--- a/lib/floe/container_runner/docker.rb
+++ b/lib/floe/container_runner/docker.rb
@@ -46,13 +46,12 @@ module Floe
       end
 
       def wait(timeout: nil, events: %i[create update delete], &block)
-        until_timestamp = Time.now.utc + timeout if timeout
+        until_timestamp = (Time.now.utc + timeout).to_i if timeout
 
         r, w = IO.pipe
 
-        pid = AwesomeSpawn.run_detached(
-          self.class::DOCKER_COMMAND, :err => :out, :out => w, :params => wait_params(until_timestamp)
-        )
+        pid = Kernel.spawn({}, self.class::DOCKER_COMMAND, *wait_params(until_timestamp), :err => :out, :out => w)
+        Process.detach(pid)
 
         w.close
 
@@ -152,8 +151,10 @@ module Floe
       end
 
       def wait_params(until_timestamp)
-        params = ["events", [:format, "{{json .}}"], [:filter, "type=container"], [:since, Time.now.utc.to_i]]
-        params << [:until, until_timestamp.to_i] if until_timestamp
+        since_timestamp = Time.now.utc.to_i
+
+        params  = ["events", "--format", "{{json .}}", "--filter", "type=container", "--since", since_timestamp.to_s]
+        params += ["--until", until_timestamp.to_s] if until_timestamp
         params
       end
 


lib/floe/container_runner/docker.rb
```diff
@@ -46,13 +46,16 @@ def cleanup(runner_context)
...
+        until_timestamp = (Time.now.utc + timeout).to_i if timeout
+        since_timestamp = Time.now.utc.to_i
+
+        args  = ["events", "--format", "{{json .}}", "--filter", "type=container", "--since", since_timestamp.to_s]
+        args += ["--until", until_timestamp.to_s] if until_timestamp
 
         r, w = IO.pipe
 
-        pid = AwesomeSpawn.run_detached(
-          self.class::DOCKER_COMMAND, :err => :out, :out => w, :params => wait_params(until_timestamp)
```
Fryguy, line 54: are we not using the wait_params method anymore? (or should it be updated/deleted)

lib/floe/container_runner/docker.rb
```diff
@@ -152,8 +151,10 @@ def run_container_params(image, env, execution_id, secrets_file)
       end
 
       def wait_params(until_timestamp)
-        params = ["events", [:format, "{{json .}}"], [:filter, "type=container"], [:since, Time.now.utc.to_i]]
-        params << [:until, until_timestamp.to_i] if until_timestamp
+        since_timestamp = Time.now.utc.to_i
+
+        params  = ["events", "--format", "{{json .}}", "--filter", "type=container", "--since", since_timestamp.to_s]
+        params += ["--until", until_timestamp.to_s] if until_timestamp
```
Fryguy, line 157: The old code did `until_timetstamp.to_i` and the new code does `to_s` (same for since_timestamp, but that one is really `to_i.to_s`) Is that correct?

lib/floe/container_runner/docker.rb
agrare: Oh yeah for testing it was easier to just throw it in the same method, moved the changes to wait_params

lib/floe/container_runner/docker.rb
```diff
@@ -46,13 +46,12 @@ def cleanup(runner_context)
       end
 
       def wait(timeout: nil, events: %i[create update delete], &block)
-        until_timestamp = Time.now.utc + timeout if timeout
+        until_timestamp = (Time.now.utc + timeout).to_i if timeout
```
agrare, line 49: NOTE moving the `.to_i` here was because I thought `.to_i.to_s` looked weird, but we need `to_i` to get the unix timestamp, and the `.to_s` is because `Kernel.spawn` needs string params.

Could alternatively use `.strftime("%s")`

Fryguy: Oh thanks - i missed that.

Fryguy (approved):

